PR #21, opened 2019-05-11
Feature/skale 1066 parallel web3 servers

sergiy-skalelabs:
Parallel web3 servers for **HTTP**, **HTTPS**, **WS** and/or **WSS** are supported by **skaled** now. Here is example how to run:
```
rm -f ./*.pem || true
openssl req -new -newkey rsa:4096 -days 365 -nodes -x509 -subj "/C=US/ST=Denial/L=Springfield/O=Dis/CN=www.example.com" -keyout key.pem -out cert.pem
skaled --aa always --web3-shutdown --config ./config0.json -d ./node --ipcpath ./ipcx -v 4 --http-port 7000 --https-port 7001 --ws-port 7002 --wss-port 7003 --web3-trace --ask 0 --bid 0 --ssl-key ./key.pem --ssl-cert ./cert.pem
```

You can see parallel servers specified above as **--http-port 7000 --https-port 7001 --ws-port 7002 --wss-port 7003**


DimaStebaev (changes requested):
Please modify `scripts/skale_build/Dockerfile` file to work with new command line options.

DimaStebaev:
I now that it is not critical but why do you use `Class::member_` instead of `Class::m_member`? Let's be consistent.

kladkogex (dismissed):

DimaStebaev (approved):

PR #28, opened 2019-05-29
Bug/skale 1086 zmq assert

dimalit:


libskale/broadcaster.cpp
```diff
@@ -161,7 +161,11 @@ void ZmqBroadcaster::startService() {
                     continue;
                 }
 
-                assert( res > 0 );
+                if ( res < 0 ) {
```
DimaStebaev, line 164: And what about `res == 0` ?

DimaStebaev (approved):

PR #31, opened 2019-05-31
Feature/skale 1103 pow gas

DimaStebaev:
This feature allows accounts without money send us pow in the gasPrice field and be granted with gas for execution of smart contracts.


sergiy-skalelabs (approved):
cool

PR #32, opened 2019-06-04
SKALE-1156 HTTP and HTTPS servers uses poll() instead of select()

sergiy-skalelabs:
Just switched to poll()


dimalit (approved):

DimaStebaev (approved):
I prefer consts to defines.

PR #52, opened 2019-07-05
Disable IPC (through geth.ipc) by default

dimalit:


cstrangedk (commented):
lgtm, defer approval to @kladkogex

kladkogex (approved):

PR #83, opened 2019-09-22
SKALE-1578 new console colorizer, trimmed/limited lo output

sergiy-skalelabs:
- New console colorizer based on **libskutils** and user everywere in **skaled**, by default it's turned off
- Auto trimmed big data values in log
- Added new **--log-value-size-limit** command line option
- By default log value size limit is set to **1024**
- Added new **--colors** and **--no-colors** command line options
- By default colors are off
**NO SPECIAL OR DEDICATED TESTS NEEDED FOR THIS PULL REQUEST**


kladkogex (changes requested):
Please specify in description if there is a test for this (or rationale why a test is not needed)

kladkogex (approved):

dimalit (commented):
I'd like to see example how to test this on "10 accounts, 4 nodes" (regression tests)

PR #111, opened 2019-10-24
SKALE-1532 crash fixes

sergiy-skalelabs:
- fixed mega deadlock in subscription mechanism
- skaled works even faster (RPC calls)

No new functions/classed added. No new tests required.


cstrangedk (approved):
lgtm

PR #113, opened 2019-10-25
SKALE-1449 dispatch domain is more error proof now

sergiy-skalelabs:
Dispatch domain is more error proof now. This means thread pool startup code is improved and handles errors and exceptions better.

No new functions/classes added. No new tests needed.


cstrangedk (approved):
lgtm

PR #118, opened 2019-11-02
SKALE-1736 fixed incorrect non-null values in the contractAddress field of TransactionReceipt JSONs

sergiy-skalelabs:
The **contractAddress** field of **Transaction Receipt** JSON must be **null** for all types of transactions but **contract deployment** ones. The **contract deployment** transaction is special because it's the only type of transaction with **to** filed set to **null**.
The **aleth** source code has the same bug not fixed yet.
The **geth** Etherereum client app behaves correctly and generates valid transaction receipts. Please take a look at line 313 [here](https://github.com/ethereum/go-ethereum/blob/master/core/types/receipt.go).
No new functions classes added.
The **simple_contract** and **contract_storage** unit tests were modified to check the **contractAddress** field correctness in transaction receipts.


DimaStebaev (changes requested):
Please add test to reproduce the bug.

DimaStebaev (approved):

PR #119, opened 2019-11-02
SKALE-978 contract code size limit customization via SKALE-specific EVM Schedule settings

sergiy-skalelabs:
To specify custom code size limit - add one of new SKALE-specific **EVM Schedules** into "params" group of **config.json** file. Here is example for unlimited code size:
        "skaleUnlimitedForkBlock": "0x0",
Here is example for 16k-bytes code size limit:
        "skale16kForkBlock": "0x0",
Here is example for 256k-bytes code size limit:
        "skale256kForkBlock": "0x0",
Here is list of supported code size limits: 16, 32, 64, 128, 256. 512. 1024 k-bytes and unlimited.

No new functions/classes added. No new tests needed. Custom **EVM Schedules** are tested by existing **testeth** app.

### DIFF
--- a/libethcore/CMakeLists.txt
+++ b/libethcore/CMakeLists.txt
@@ -1,6 +1,13 @@
-file(GLOB sources "*.cpp" "*.h")
+file( GLOB sources "*.cpp" "*.h" )
 
-add_library(ethcore ${sources})
+add_library( ethcore ${sources} )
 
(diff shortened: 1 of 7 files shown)


dimalit (approved):
ok, but i'm afraid this could be introduction of unnecessary extra complexity

PR #129, opened 2019-11-20
Feature/skale 1790 integrate snapshot hashes into skaled

olehnikolaiev:
medium risk
add new snapshots downloading functionality - starting from now a node should first get snapshot's hash from all other nodes and takes the one that simillar to at least 2/3 of others. after that node randomly chooses a node to download snapshot from and checks snapshot hash after downloading.
test for this feature here -> https://github.com/skalenetwork/skale-node-tests/blob/master/sktest_snapshot.py


kladkogex (approved):

kladkogex:
Build fails

PR #140, opened 2019-12-05
Bug/skale 1854 default queue leaks

sergiy-skalelabs:
- HTTP(S) queues count is limited and configurable
- Servers support and understand shutdown mode
- Arrived message payloads ignored in shutdown mode
- new WS(S) connections rejected in shutdown mode
- fixed bad WS peer reference capture in lambda functions causing crash on shutdown
- no new functions/classes added
- no new tests required


kladkogex (dismissed):

cstrangedk (commented):
lgtm

dimalit (approved):

PR #156, opened 2020-01-03
SKALE-1930 blocks_N.db is also snapshotted as prices_N.db

dimalit:
To get rid of "got null" messge


DimaStebaev (approved):

dimalit:
merged with SKALE-1918-blocks-rotation

PR #228, opened 2020-03-24
SKALE-2352 Fixed skaled build errors on clang and Mac OSX

sergiy-skalelabs:
There were build errors on latest Apple's **clang** and Mac OSX. Once upon a time, a very early morning I was in a good mood to fix this. No new functions/classes added. No new tests required.
In details, the **MSIZE** symbol is defined somewhere in **EVM** code and names some instruction. The same **MSIZE** symbol is defined as integer code in network related headers of Apple SDK. I would recommend to rename **MSIZE** in **EVM** to avoid same problems in the future.

### DIFF
--- a/libskutils/include/skutils/http.h
+++ b/libskutils/include/skutils/http.h
@@ -3,7 +3,8 @@
 #if ( !defined SKUTILS_HTTP_H )
 #define SKUTILS_HTTP_H 1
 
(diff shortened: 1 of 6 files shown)


kladkogex (approved):

olehnikolaiev (commented):
great! now it works!

PR #283, opened 2020-05-29
SKALE-2589 fixed IMA token message verification

sergiy-skalelabs:
…disassembled incorrectly


kladkogex (approved):

payvint (approved):
Remember this, nice catch!

PR #291, opened 2020-06-13
Add snapshot and JSON-RPC documentation

cstrangedk:


DimaStebaev (approved):
JSON RPC is OK.
It's better if @dimalit reviews snapshots docs.

dimalit:
config parameter is called snapshotIntervalMs. is it ok to name it just snapshotInterval?

cstrangedk:
> config parameter is called snapshotIntervalMs. is it ok to name it just snapshotInterval?

`snapshotIntervalMs` more informatively indicates units in milliseconds. I think that this is a more useful config variable name.

PR #306, opened 2020-07-10
Bug/skale 2755 state root

olehnikolaiev:
low risk
fix stateRoot issue, update config


libskale/SnapshotManager.cpp
```diff
@@ -272,14 +272,36 @@ void SnapshotManager::leaveNLastSnapshots( unsigned n ) {
     }      // for
 }
 
+std::pair< int, int > SnapshotManager::getLatestSnasphot() const {
```
dimalit, line 275: why don't you just use directory name (it is block number) instead of time stamp?

dimalit (approved):

PR #333, opened 2020-08-20
bug/SKALE-3135-shutdown-can-be-better 

sergiy-skalelabs:
- Added several loop termination statements detecting exit requested situation
- Improved Ethereum Client shutdown actions sequence handling
- Shutdown **JSON RPC** interfaces earlier. let other subsystems shutdown easier
- Added signal handling on startup when consensus waits for other **un-started** S-chain nodes
- **No new** functions/classes added
- **No new** tests needed


libethereum/BlockQueue.cpp
```diff
@@ -49,11 +49,16 @@ BlockQueue::BlockQueue() {
     // Allow some room for other activity
     unsigned verifierThreads = 1;  // needed for JsonRpcTests (real mining)  // std::max(
                                    // thread::hardware_concurrency(), 3U ) - 2U;
-    for ( unsigned i = 0; i < verifierThreads; ++i )
+    for ( unsigned i = 0; i < verifierThreads; ++i ) {
+        if ( this->m_deleting )
+            return;
         m_verifiers.emplace_back( [=]() {
+            if ( this->m_deleting )
```
dimalit, line 56: This is not needed, as verifierBody has this check exactly in its while() loop
```
void BlockQueue::verifierBody() try {
    while ( !m_deleting ) {
```

skaled/main.cpp
```diff
@@ -313,6 +313,9 @@ get_machine_ip_addresses_6() {  // first-interface name, second-address
     return listIfaceInfos6;
 }
 
+static std::unique_ptr< Client > g_client;
+unique_ptr< ModularServer<> > g_jsonrpcIpcServer;
```
dimalit, line 318: Why are they static? I don't see any reason why they cannot be main()'s local vars

skaled/main.cpp
```diff
@@ -2239,17 +2250,23 @@ int main( int argc, char** argv ) try {
...
-    if ( jsonrpcIpcServer.get() )
-        jsonrpcIpcServer->StopListening();
+    }
+    if ( g_jsonrpcIpcServer.get() ) {
+        g_jsonrpcIpcServer->StopListening();
+        g_jsonrpcIpcServer.release();
+    }
+    if ( g_client ) {
+        g_client->stopWorking();
+        g_client.release();
```
dimalit: Attention! release() does not delete object under unique_ptr! To do this, we need reset(nullptr)!!!
(here and in other places)

libethereum/Client.cpp
```diff
@@ -119,9 +119,23 @@ Client::Client( ChainParams const& _params, int _networkID,
 }
 
 Client::~Client() {
+    stopWorking();
+}
+
+void Client::stopWorking() {
```
dimalit, line 125: I just realized that first thing that we need to do in Client's stopWorking - is to call m_skaleHost->stopWorking(). This is very important - because it will ensure that we are not in the middle of block importing! (to archieve this - SkaleHost waits for m_consensusWorkingMutex to become available). Also, important that join()'ing of m_snapshotHashComputing was done after this too.

libethereum/Client.cpp
```diff
@@ -119,9 +119,23 @@ Client::Client( ChainParams const& _params, int _networkID,
...
+    Worker::stopWorking();
+
     if ( m_snapshotHashComputing != nullptr )
         m_snapshotHashComputing->join();
 
+    if ( m_snapshotHashComputing != nullptr ) {
+        try {
+            if ( m_snapshotHashComputing->joinable() )
```
dimalit, line 136: We don't need here neither `if` no `try` - because the only other join() with this thread is done in Client::importTransactionsAsBlock, and that function is executed under mutex which is waited to be released in SkaleHost::stopWorking() (m_consensusWorkingMutex)

libethereum/Client.cpp
```diff
@@ -132,17 +146,32 @@ Client::~Client() {
...
 
     m_signalled.notify_all();  // to wake up the thread from Client::doWork()
-    stopWorking();
 
     m_tq.HandleDestruction();  // l_sergiy: destroy transaction queue earlier
     m_bq.stop();               // l_sergiy: added to stop block queue processing
 
-    terminate();
+    m_bc.close();
+    std::cout << cc::success( "Blockchain is closed" ) << "\n";
```
dimalit: Better write to log - and not to stdout

libethereum/Client.cpp
```diff
@@ -132,17 +146,32 @@ Client::~Client() {
...
+        std::cout << cc::success( "Deleted lock file " )
+                  << cc::p( boost::filesystem::canonical( m_dbPath ).string() +
+                            std::string( "/skaled.lock" ) )
+                  << "\n";
+    else
+        std::cout << cc::fatal( "ATTENTION:" ) << " " << cc::error( "Deleted lock file " )
+                  << cc::p( boost::filesystem::canonical( m_dbPath ).string() +
+                            std::string( "/skaled.lock" ) )
+                  << cc::error( " after foreceful exit" ) << "\n";
+    std::cout.flush();
```
dimalit: If write to log (see above) then here we also need to flush log, not cout.

dimalit (changes requested):
There are important changes that are requested! Also, I didn't look at libskutils/* - as I'm not familiar with those lands.

libethereum/BlockQueue.cpp
sergiy-skalelabs: double check is a problem? it allows to handle situation earlier

libethereum/Client.cpp
sergiy-skalelabs: debugging of many simiar joins shows it's needed

libethereum/Client.cpp
sergiy-skalelabs: This can be removed after resolving problems with shutdown

libethereum/Client.cpp
sergiy-skalelabs: This can be removed after resolving problems with shutdown

skaled/main.cpp
sergiy-skalelabs: because they need to be called from c-callback handling signals - no C++ capture in this callback

libethereum/BlockQueue.cpp
dimalit: Not a problem, but too many redundant checks make reading code uncomfortable. In this case two checks of m_deleting have only on line between them: `setThreadName`!

libethereum/Client.cpp
dimalit: then it means that we had broken logic elsewhere and need to fix in that place - and not in this!?

libethereum/Client.cpp
sergiy-skalelabs: resolved

libethereum/Client.cpp
sergiy-skalelabs: resolved

libethereum/Client.cpp
sergiy-skalelabs: I trust you here.... resolved... nice catch!

skaled/main.cpp
sergiy-skalelabs: And g_jsonrpcIpcServer too... resolved... nice catch!

libethereum/Client.cpp
```diff
@@ -119,30 +119,56 @@ Client::Client( ChainParams const& _params, int _networkID,
...
     m_tq.HandleDestruction();  // l_sergiy: destroy transaction queue earlier
     m_bq.stop();               // l_sergiy: added to stop block queue processing
 
-    terminate();
+    m_bc.close();
+    LOG( m_logger ) << cc::success( "Blockchain is closed" );
 
-    if ( !m_skaleHost || m_skaleHost->exitedForcefully() == false )
+    bool isForcefulExit =
+        ( !m_skaleHost || m_skaleHost->exitedForcefully() == false ) ? false : true;
```
dimalit, line 154:
```
bool isForcefulExit =
        m_skaleHost && m_skaleHost->exitedForcefully();
```
(equivalent)

dimalit (approved):

PR #338, opened 2020-08-27
Make format faster

dimalit:
Guys, do you agree with this variant of regex in clang-format.cmake?


dimalit:
Looks like it has no effect

PR #344, opened 2020-09-02
feature/SKALE-2814 partial catch up

sergiy-skalelabs:
Partial catch up, DB safety issues.
If **skaled** crashes after some transactions executed and their **state data** was saved into **level DB** (i.e. block was not finished) then next startup these transactions are skipped and only rest of transactions are executed to finish block mining.

**NOTICE:** the new `Contract Events Integrity Test` was used to verify partial catch-up data, you can find it here:
https://github.com/skalenetwork/SkaleExperimental/tree/master/skaled-tests/test-event-integrity

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -482,7 +482,8 @@ size_t Client::importTransactionsAsBlock(
         // begin, detect partially executed block
         bool bIsPartial = false;
         size_t cntAll = _transactions.size();
-        size_t cntEpected = cntAll;
+        size_t cntExpected = cntAll;
+        size_t cntMissing = 0;
         Transactions vecPassed, vecMissing;
         dev::h256 shaLastTx = m_state.safeLastExecutedTransactionHash();
         for ( const Transaction& txWalk : _transactions ) {
@@ -494,12 +495,13 @@ size_t Client::importTransactionsAsBlock(
                 if ( shaWalk == shaLastTx ) {
                     bIsPartial = true;
                     size_t cntPassed = vecPassed.size();
-                    size_t cntMissing = cntAll - cntPassed;
-                    cntEpected = cntMissing;
-                    LOG( m_logger ) << cc::warn( "WARNING: found partially executed block, have " )
-                                    << cc::size10( cntAll ) << cc::error( " transaction(s), " )
-                                    << cc::size10( cntPassed ) << cc::error( "passed, " )
-                                    << cc::size10( cntMissing ) << cc::error( " missing" );
+                    cntMissing = cntAll - cntPassed;
+                    cntExpected = cntMissing;
+                    LOG( m_logger ) << cc::fatal( "PARTIAL CATCHUP DETECTED:" )
+                                    << cc::warn( " found partially executed block, have " )
+                                    << cc::size10( cntAll ) << cc::warn( " transaction(s), " )
+                                    << cc::size10( cntPassed ) << cc::warn( " passed, " )
+                                    << cc::size10( cntMissing ) << cc::warn( " missing" );
                     LOG( m_logger ).flush();
                 }
             }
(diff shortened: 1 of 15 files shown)


libethereum/Client.cpp
```diff
@@ -226,6 +226,8 @@ void Client::init( WithExisting _forceAction, u256 _networkId ) {
     if ( m_dbPath.size() )
         Defaults::setDBPath( m_dbPath );
 
+    bc().m_stateDB = m_state.db();
+
```
dimalit, line 239: Why was this necessary?

libskale/State.cpp
```diff
@@ -817,7 +824,19 @@ std::pair< ExecutionResult, TransactionReceipt > State::execute( EnvInfo const&
 
         removeEmptyAccounts = _envInfo.number() >= _sealEngine.chainParams().EIP158ForkBlock;
         commit( removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts :
-                                      State::CommitBehaviour::KeepEmptyAccounts );
+                                      State::CommitBehaviour::KeepEmptyAccounts,
+            [&]( std::shared_ptr< dev::db::DatabaseFace > /*db*/,
+                std::unique_ptr< dev::db::WriteBatchFace >& writeBatch ) {
+                if ( isSaveLastTxHash ) {
```
dimalit, line 846: Why this cannot be always true?

libskale/State.cpp
```diff
@@ -348,7 +355,7 @@ void State::commit( CommitBehaviour _commitBehaviour ) {
             }
         }
         m_db_ptr->updateStorageUsage( totalStorageUsed_ );
-        m_db_ptr->commit();
+        m_db_ptr->commit( fn_pre_commit );
```
dimalit, line 374: Why can't we pass last transaction hash to OverlayDB in the same way as we pass totalStorageUsed_?
Do we really need this callback mechanism?

dimalit (commented):
Where does this use new ConsensusEngine::getBlock() API call?
(see also other comments)

libethereum/Client.cpp
sergiy-skalelabs: This is initialization of variable which must be non-null smart pointer earlier

libskale/State.cpp
sergiy-skalelabs: I am just providing new and old versions of methods at once
isSaveLastTxHash = true -> my new version needed for partial catch up

libskale/State.cpp
sergiy-skalelabs: We cannot pass last transaction hash. This is not what is needed here. We need to do operations which modify batch actions created in memory. This is same as transactions in classic SQL. Function handler to add level DB batch actions looks like the only way

libskale/State.cpp
dimalit: So can we just remove this parameter and work as if it's value is true?

libskale/State.cpp
dimalit: Yes, of course we need it in batch. But what's the problem to have "last transaction hash" field in OverlayDB class and have it put this field writing to DB into batch?

libethereum/Client.cpp
dimalit: I meant this:
`bc().m_stateDB = m_state.db();`

This looks like intruding into BlockChain class' private parts..

libskale/State.cpp
sergiy-skalelabs: yes, right, but I need it initialized

libethereum/Client.cpp
sergiy-skalelabs: yes, right, but I need it initialized

libskale/State.cpp
sergiy-skalelabs: State::execute() is invoked several times in code and I am sure I need new isSaveLastTxHash parameter to be true only in one place, and not sure about other places, so isSaveLastTxHash is too early to remove

dimalit (commented):
1. I'm not yet sure, that in current implementation we won't get duplicated receipts.
2. Post-commit callback can be eliminated from State - will do it tomorrow
3. Still don't see where particular bloom filter can be corrupted. Will look more

sergiy-skalelabs:
> 1. _I'm not yet sure, that in current implementation we won't get duplicated receipts._
**ANSWER:** we cannot get duplicated receipts, everything is detected at right place
> 2. _Post-commit callback can be eliminated from State - will do it tomorrow_
**ANSWER:** need to discuss this
> 3. _Still don't see where particular bloom filter can be corrupted. Will look more_
**ANSWER:** it's not corrupted, it's absent time to time in log search results, but saved into extras DB always

kladkogex (approved):

PR #353, opened 2020-09-22
Feature/skale 2814 partial catch up 2

dimalit:
Make OverlayDB fully manage lastTransactionHash as part of state.
Also some small changes in the logic of tracking of storage size.


dimalit:
Closing as outdated

PR #360, opened 2020-10-12
SKALE-3373 now eth_getTransactionReceipt returns status as 0x string

sergiy-skalelabs:
- now `eth_getTransactionReceipt` returns `status` field of response JSON as `0x` string
- no new tests needed

### DIFF
--- a/libdevcore/CommonIO.h
+++ b/libdevcore/CommonIO.h
@@ -230,4 +230,26 @@ inline std::string toString< uint8_t >( uint8_t const& _u ) {
     o << static_cast< uint16_t >( _u );
     return o.str();
 }
+
+template < class T >
+inline std::string toString0x( T const& _val ) {
+    std::ostringstream o;
+    o << "0x" << std::uppercase << std::setfill( '0' ) << std::setw( 1 ) << std::hex << _val;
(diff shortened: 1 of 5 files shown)


dimalit (dismissed):
How can we be sure that we won't have similar problems with other calls?

sergiy-skalelabs:
we cannot be sure about other JSON RPC APIs
no insurance at all here
this was something specific for Aleth only
we only can fix everything similar one by one in the future

dimalit:
Looks like in our tests we expect this method to return result without 0x - so, now tests fail.
E.g.
```2020-10-13T06:18:45.6890179Z /home/runner/work/skaled/skaled/test/unittests/libweb3jsonrpc/jsonrpc.cpp(1127): [4;31;49mfatal error: in "JsonRpcSuite/eth_sendRawTransaction_gasLimitExceeded": critical check receipt["status"] == string( "0" ) has failed ["0x0" != 0][0;39;49m```

sergiy-skalelabs:
- ticket tells me to add 0x
- other Ethereum clients have 0x-hex-format here
- looks like that test is wrong
)))

dimalit (approved):

PR #386, opened 2020-11-16
Updated consensus Fixes for SKALE-3545 SKALE-3546 SKALE-3553

kladkogex:


dimalit:
1 Remove busy wait when waiting for sgx server
2 Do not crash on assert in some places in conensus

PR #387, opened 2020-11-17
SKALE-3436 Download snapshot from random

dimalit:
should be OK, but don't merge without testing, please
also, there's a question if we can call srand() in main.cpp
(can potentially conflict with such call in consensus!)


kladkogex (dismissed):

kladkogex:
Approved

PR #389, opened 2020-11-23
SKALE-3554 fix IMA transfer for ERC20 and ERC721 tokens

sergiy-skalelabs:
- correct data byte order of IMA messages, had problem with incorrect BLS signatures for ERC20 and ERC721 tokens
- no new tests needed


payvint (dismissed):
Nice fix!!

payvint (dismissed):
Nice fix!

dimalit (approved):

PR #443, opened 2021-01-27
SKALE-3802 fix calculating hash

olehnikolaiev:


dimalit (changes requested):
Need to iterate through subdirs in blocks and prices too

dimalit (dismissed):

PR #448, opened 2021-01-29
Revert "SKALE-3802 fix calculating hash"

dimalit:
Reverts skalenetwork/skaled#443


dimalit:
Already fixed in another PR.

PR #509, opened 2021-03-05
SKALE-3962 cleanup before starting from snapshot

olehnikolaiev:


dimalit (commented):
I think with these changes we don't need check for same snapshot presence anymore. BUT need to carefully check case latest snpshot == 0. I will take a look.

dimalit (commented):
UPD If we crashed, and it's just docker restarting us - need to check if can just safely run without any download!

dimalit (changes requested):
I blocked PR until we implement mentioned above updates.

sergiy-skalelabs (dismissed):

sergiy-skalelabs (dismissed):

PR #528, opened 2021-03-22
SKALE-2512 finalize switch to libWebSockets 4.1

sergiy-skalelabs:
- switched to manual OpenSSL initialization to avoid conflicts between multiple servers
- no new tests needed


sergiy-skalelabs:
will re-create this PR

PR #531, opened 2021-03-23
SKALE-2512 switched to manual SSL init

sergiy-skalelabs:
- Update for newest libWebSockets 4.1 integration.
- Open SSL is initialized and de-initialized exclusively in libWebSockets 4.1 with no care about external developer needs. This leads to incorrect Open SSL shutdown in skaled threads and random wss connection crash on connection close. Fixed with manual Open SSL initialization.


dimalit (approved):
LGTM

PR #545, opened 2021-03-29
Do not closeSend() and closeReceive()

dimalit:


dimalit:
Implemented in anothe way (no poll())

PR #681, opened 2021-06-14
SKALE-4280 add snapshot time to debug api

olehnikolaiev:


dimalit (approved):
I think we need to refactor this later (too messy Client class), but for quick solution - ok

PR #687, opened 2021-06-15
Add ability to build with externally-set -O2 flag

dimalit:


dimalit:
Not relevant, as we already have Debug/Release/RelWithDebugInfo builds

PR #700, opened 2021-06-23
Use -O2 -g in container build

dimalit:


dimalit:
Not relevant, as we already have Debug/Release/RelWithDebugInfo builds

PR #751, opened 2021-08-13
SKALE-4495 iterative logs search

sergiy-skalelabs:
- iterative logs search for IMA message validation
- use ZMQ to BLS-sign IMA mesages


sergiy-skalelabs:
will re-create it without zmq

PR #792, opened 2021-09-07
Added --sgx-url-no-zmq command line argument

sergiy-skalelabs:
Added `--sgx-url-no-zmq` command line argument which makes **IMA** related code work with **SGX** via **HTTPS** instead of **ZMQ**, by default this is turned off and **ZMQ** is used. This PR is for testing purposes of ALL SKALE test.


sergiy-skalelabs:
no longer needed

PR #854, opened 2021-10-25
Add precompile info for 3.8

cstrangedk:


docs/modules/ROOT/pages/random-number-generator.adoc
```diff
@@ -0,0 +1,43 @@
...
+            if iszero(staticcall(gas, 0x18, 0, 0, start_addr, 32)) {
+              invalid()
+            }
+            addr := mload(freemem)
+        }
+    }
+}
+----
+
+Running getRandom() from Remix several times will return the following example output:
```
DmytroNazarenko: I think need to append with `output for each of the new blocks:`

docs/modules/ROOT/pages/precompiles.adoc
```diff
@@ -0,0 +1,79 @@
+= SKALE Chain Precompiles
+
+Each SKALE Chain comes with a set of precompiled contracts in addition to the precompiles included with Ethereum (i.e. `ecrecover`, `sha256`, `modexp`, ...).
```
DmytroNazarenko: Need to add some info about `restrictAccess` option - to protect precompiled from external calls. Only specific address could call marked precopmileds. This option is used for filestorage precompileds (0x0B, 0x0C, 0x0E,0x0F, 0x10). Other are free to call. Example:
```
                "startingBlock": "0x0",
                "restrictAccess": [
                    "0xD3002000000000000000000000000000000000D3"
                ]
```

DmytroNazarenko (changes requested):

docs/modules/ROOT/pages/precompiles.adoc
cstrangedk: added new column for clarification

DmytroNazarenko (approved):

PR #882, opened 2021-11-08
3.7.5 beta.5 hotfix 1

dimalit:
1 Fix crash on JSON-RPC
2 Fix block rotation
3 Output current DB usage to -v 3 log


dimalit:
Discard this as this used wrong beta (5 instead of 4)

PR #883, opened 2021-11-08
3.7.5 beta.4 hotfix 1

dimalit:
1 Fix crash on JSON-RPC
2 Fix block rotation
3 Output current DB usage to -v 3 log
4 Fix EIP for JSON-RPC


dimalit:
Not relevant

PR #945, opened 2022-01-10
Feature/skale 1263 game mode

DmytroNazarenko:
Adding MultiTransaction mode


libethereum/SkaleHost.cpp
```diff
@@ -376,9 +376,11 @@ ConsensusExtFace::transactions_vector SkaleHost::pendingTransactions(
 
             if ( tx.verifiedOn < m_lastBlockWithBornTransactions )
                 try {
```
dimalit: During normal operation, transactions with future nonces will not get here! So we don't need to set mtmEnabled here!
(or probably can even test nonce with assert)

libethereum/Client.cpp
```diff
@@ -1163,11 +1163,18 @@ h256 Client::importTransaction( Transaction const& _t ) {
...
         bc().number() ? this->blockInfo( bc().currentHash() ) : bc().genesis(), state,
-        *bc().sealEngine(), 0, gasBidPrice );
+        *bc().sealEngine(), 0, gasBidPrice, multitransactionMode );
+
+    ImportResult res;
+    if ( multitransactionMode && state.getNonce( _t.sender() ) < _t.nonce() ) {
+        res = m_tq.import( _t, IfDropped::Ignore, true );
+    } else {
+        res = m_tq.import( _t );
+    }
```
dimalit, line 1178: why would we need two separate ways to import "current" and "future" transaction? can't the transaction queue determine by itself where to put new txn? (and in fact it first puts every new txn into "current" queue and then conditionally moves it into the "future" queue!)

dimalit (changes requested):

dimalit (approved):

PR #964, opened 2022-01-21
Enhancement/skale 4718 oracle api

DmytroNazarenko:
Add `oracle_submitRequest` and `oracle_checkResult` API


libethereum/Client.h
```diff
@@ -301,6 +301,9 @@ class Client : public ClientBase, protected Worker {
         return this->snapshot_hash_calculation_time_ms;
     }
 
+    uint64_t submitOracleRequest( const string& _spec, string& _receipt );
+    uint64_t checkOracleResult( string& _receipt, string& _result );
+
```
dimalit, line 306: Probably, here we need const string& as checkOracleResult's 1st parameter

libethereum/ConsensusStub.h
```diff
@@ -39,7 +39,8 @@ class ConsensusStub : private dev::Worker, public ConsensusInterface {
 public:
     ConsensusStub( ConsensusExtFace& _extFace, uint64_t _lastCommittedBlockID, u256 _stateRoot );
     ~ConsensusStub() override;
-    void parseFullConfigAndCreateNode( const std::string& _jsonConfig ) override;
+    void parseFullConfigAndCreateNode(
+        const std::string& _jsonConfig, const string& _gethURL ) override;
     void startAll() override;
```
dimalit, line 44: Probably, here we could give it default value of string("")

kladkogex (approved):

dimalit (approved):

PR #1062, opened 2022-04-06
Bug/skale 5059 limit number of filters per ip

### LINKED ISSUE #1043: Limit number of filters pe IP address
Limit number of filters per IP address to a value settable by schain owner.

┆Issue is synchronized with this [Jira Task](https://skalelabs.atlassian.net/browse/SKALE-5059)

dimalit:


dimalit (commented):
I see 2 drawbacks of such approach:
1) UnDoS protection is coupled with the implementation (of logs/filter functionality). Architecturally more correct is to separate these two.
2) It unnecessarily introduces ("manual") parsing of filter/logs requests through rapidjson (and not through plain libjsonrpccpp as before. This can can introduce new possible bugs (in this new code).
2a) As a side note I should mention that currently rapidjson-related part in SkaleServerOverride is implemented in a very "workaround" way and needs re-architecture. Growing it further with more calls would complicate things.

As a solution, I'd propose to implement this protection as an added layer in SkaleServerOverride class. It already has unddos protection - so this addition looks consistent. It can be implemented through something similar to mod_rewrite of Apache - i.e. scan request with simple regexp or plain c++ code and do appropriate filtering.

PR #1082, opened 2022-04-27
skaled/beta` can be built on Ununtu 22.04

sergiy-skalelabs:
`skaled`/`beta` can be built on **Ununtu 22.04**


sergiy-skalelabs:
wrong source branch

PR #1096, opened 2022-05-17
Bugfix/ticket 1091 fixed sgx url with path usage in stable

sergiy-skalelabs:
Fixed problem causing skaled loosing connectivity with SGX wallet if SGX URL contains path


sergiy-skalelabs:
Wrong source branch

PR #1247, opened 2022-12-07
Global UnDDoS limit settings affecting all connections(in syncnode_develop)

### LINKED ISSUE #1241: Need to provide ddos protection for all IPs
Currently our DDOS protection handles and limits calls from one IP. We need to have an ability to handle overall number of calls from all IPs. Will help us to have readable errors and strict limits for calls. 

Need to add this to "unddos" section of skaled config

sergiy-skalelabs:
Implemented global **UnDDoS** limit settings affecting all connections on all network protocols(in `syncnode_develop`)


sergiy-skalelabs:
Will re-create PR with less code formatting

PR #1275, opened 2022-12-19
Fixed global UnDDoS computations(in develop)

### LINKED ISSUE #1241: Need to provide ddos protection for all IPs

sergiy-skalelabs:
Fixed global UnDDoS computations(in develop)


DmytroNazarenko:
Will be merged with syncnode_develop

PR #1283, opened 2022-12-20
# 1270 Remove consensusdb from snapshots

### LINKED ISSUE #1270: Remove consensus DB from snapshot calculation
We should decrease disk usage during/after snapshot creation on skale node

Related to https://github.com/skalenetwork/internal-support/issues/575

kladkogex:
Fixes #1270


sergiy-skalelabs (dismissed):

DmytroNazarenko (changes requested):
Fix code format, please

PR #1322, opened 2023-01-06
Beta without JSON validation

dimalit:


kladkogex (approved):

sergiy-skalelabs (approved):

dimalit:
Fails probably because of bug in GitHub

PR #1470, opened 2023-03-22
Add clang-format-12 install to README

### LINKED ISSUE #1467: Transition skaled to clang-format 11
It's latest available in Ubuntu 20.04, and it is available in Ubuntu 22.04

UPD: decided to use clang-format 11, because clang 12 working differently on 20.04 & 22.04

dimalit:


dimalit:
Merged with another PR

PR #1507, opened 2023-04-28
731 memory leaks

kladkogex:
Was tested using existing consensus test

Memory growth in consensus stopped.

Consensus now take 200MByte max under large block load


sergiy-skalelabs (dismissed):

dimalit:
UPD Consensus seems to have no infuence on memory growth. But limiting transaction queue to 100 txns - allowed to limit RSS to 1Gb instead of 7Gb on 50-kb transactions

DmytroNazarenko (approved):

PR #1514, opened 2023-05-04
1425 Threading rules

### LINKED ISSUE #1425: Establish threading rules in skaled-consensus interaction
1. Current implementation is based on the assumption that consensus always calls pendingTransactions and createBlock from the same thread.This is not the case for very long time already.
2. Threading rules for startAll/bootstrapAll should be defined.
3. Threading rules for consensus and skaled termination should be defined. When these functions can be called? From which threads they can be called?
4. In ConsensusExtFace::terminateApplication() should consensus' exitGracefully() be also called?
5. Evaluate this statement: `Consensus guarantees that it will not do anything for a particular block ID, until pendingTransactions(...)
     for this block id returns.` (ConsensusInterface.h)
6. Consensus should have 2 states: 'created' and 'started'. After 1st we can allow RPC. See https://github.com/skalenetwork/internal-support/issues/753

Need to decide these rules, document and follow them.

kladkogex:


dimalit (approved):

dimalit (changes requested):
This is WRONG branch!
Correct one is enhancement/SKALED-1425-threading-rules

PR #1540, opened 2023-06-08
Feature/1506 run rpc without 2/3 nodes

### LINKED ISSUE #1506: Need ability to run RPC API when no 2/3 of nodes
When skaled starts when there are no 2/3 of nodes, it waits for them and doesn't allow others to send JSON RPC requests to it. In some situations this behavior can prevent chain from repairing. So, we need to start JSON-RPC before consensus

olehnikolaiev:
tests are here https://github.com/skalenetwork/skale-node-tests/pull/6


kladkogex:
Lets discuss on Monday )

kladkogex:
I added logic to consensus so it is does not wait for EMPTY_BLOCK_TIMEOUT during bootstrap all 
It is cleaner

kladkogex (changes requested):
I moved the logic to setEmptyBlockInterval to 50 ms initially to consensus, 
it is cleaner this way.

Please modify the pull request)

I think it may be not needed anymore

DmytroNazarenko (approved):

sergiy-skalelabs (approved):

kladkogex (approved):

PR #1541, opened 2023-06-13
1450 create diffPath after cleanup if needed

### LINKED ISSUE #1450: Be able to catchup from 0 block
**What?**
We need to catchup 0 block in the process of catchup. Currently we run catchup from 1st block and for 0 block we are using genesis from config

**Why?**
To be able to catchup genesis block from another node and do not be dependent from predeployeds in the config. These can help us to avoid STATE_ROOT because of different genesis block

**Who?**
This feature will help engineering team and archive node maintainers. It will help to simplify maintaining and deployment process of the archive node

olehnikolaiev:
fixes https://github.com/skalenetwork/skaled/issues/1450
in some rare cases diffPath was deleted and not created back that caused errors during snapshot downloading phase


dimalit (approved):

sergiy-skalelabs (approved):

DmytroNazarenko:
Presented by @dimalit @olehnikolaiev:
- Passing empty path to shared-space in docker-lvmpy, creating path by skaled

DmytroNazarenko (approved):

PR #1677, opened 2023-09-28
Bug/bug 1198 dont ignore params eth block number

### LINKED ISSUE #1198: Skaled ignore params field for eth_blockNumber method
**Versions**
Skaled: schain:3.14.15-stable.2

**Environment**
20 active nodes on the network
2 schains up and running 

**Steps to Reproduce/Description**
1. send eth_blockNumber request with params to skaled 
2. check response 

**Expected behavior**
skaled should return an error 
```""error":{"code":-32602,"message":"too many arguments, want at most 0"}"```

**Actual result** 
skaled return the latest block number without any errors 
```{"id":1,"jsonrpc":"2.0","result":"0xcbce7"}``` 

example request 
```
curl http://sample-endpoint-name.network.quiknode.pro/token-goes-here/ \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"eth_blockNumber","params":["0x40ee315dbee7c25d0ca4265d75b545cc87529edc5590bc2324eef0a706ab9c14"],"id":1,"jsonrpc":"2.0"}'
```

### PR DESCRIPTION
**Pull request by olehnikolaiev**
fixes #1198

### DIFF
--- a/libweb3jsonrpc/Eth.cpp
+++ b/libweb3jsonrpc/Eth.cpp
@@ -110,7 +110,10 @@ Json::Value Eth::eth_accounts() {
     return toJson( m_ethAccounts.allAccounts() );
 }
 
-string Eth::eth_blockNumber() {
+string Eth::eth_blockNumber( const Json::Value& request ) {
+    if ( !request.empty() ) {
+        BOOST_THROW_EXCEPTION( JsonRpcException( Errors::ERROR_RPC_INVALID_PARAMS ) );
+    }
     return toJS( client()->number() );
 }
 
--- a/libweb3jsonrpc/Eth.h
+++ b/libweb3jsonrpc/Eth.h
@@ -77,7 +77,7 @@ public:
     virtual bool eth_mining() override;
     virtual std::string eth_gasPrice() override;
     virtual Json::Value eth_accounts() override;
-    virtual std::string eth_blockNumber() override;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: libweb3jsonrpc/Eth.cpp [hunk 1]
```diff
@@ -110,7 +110,10 @@ Json::Value Eth::eth_accounts() {
     return toJson( m_ethAccounts.allAccounts() );
 }
 
-string Eth::eth_blockNumber() {
+string Eth::eth_blockNumber( const Json::Value& request ) {
+    if ( !request.empty() ) {
+        BOOST_THROW_EXCEPTION( JsonRpcException( Errors::ERROR_RPC_INVALID_PARAMS ) );
```
DmytroNazarenko, line 115: We need to return the same message as Ethereum does - `{"code":-32602,"message":"too many arguments, want at most 0"}`

DmytroNazarenko (changes requested):

dimalit (approved):

libweb3jsonrpc/Eth.cpp
olehnikolaiev: this is our standard error message for all json-rpc methods. should we change all of them?

DmytroNazarenko: Let's raise a ticket about compatibility of error messages with EVM's and change them in separate PR's, then

DmytroNazarenko (approved):

PR #1687, opened 2023-10-06
Fix for duplicatiing transactions (IS-869)

dimalit:
1. Skip invalid transactions from block, starting from "patch timestamp"
2. Ignore them in historic build, if they are present in block

Fixes https://github.com/skalenetwork/internal-support/issues/869

Tests:
1. test/unittests/libethereum/SkaleHost.cpp tests how various kinds of invalid transactions are skipped from block (both for enabled patch and for disabled patch)
2. test/unittests/libweb3jsonrpc/jsonrpc.cpp -> skip_invalid_transactions tests affected JSON-RPC calls on historic node

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -55,6 +55,7 @@
 #include <libskale/ContractStorageZeroValuePatch.h>
 #include <libskale/POWCheckPatch.h>
 #include <libskale/RevertableFSPatch.h>
+#include <libskale/SkipInvalidTransactionsPatch.h>
 #include <libskale/State.h>
 #include <libskale/StorageDestructionPatch.h>
 #include <libskale/TotalStorageUsedPatch.h>
@@ -162,6 +163,7 @@ Client::Client( ChainParams const& _params, int _networkID,
     RevertableFSPatch::setTimestamp( chainParams().sChain.revertableFSPatchTimestamp );
     StorageDestructionPatch::setTimestamp( chainParams().sChain.storageDestructionPatchTimestamp );
     POWCheckPatch::setTimestamp( chainParams().sChain.powCheckPatchTimestamp );
+    SkipInvalidTransactionsPatch::init( this->chainParams() );
 }
 
 
@@ -654,7 +656,7 @@ size_t Client::syncTransactions(
     RevertableFSPatch::lastBlockTimestamp = blockChain().info().timestamp();
     StorageDestructionPatch::lastBlockTimestamp = blockChain().info().timestamp();
     POWCheckPatch::lastBlockTimestamp = blockChain().info().timestamp();
-
+    SkipInvalidTransactionsPatch::setLastBlock( blockChain().info() );
 
     DEV_WRITE_GUARDED( x_working ) {
         assert( !m_working.isSealed() );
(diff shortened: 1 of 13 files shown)


kladkogex (changes requested):
Please specify in detail how the fix was tested.

Also please add comments to the code describing the fix

libskale/SkipInvalidTransactionsPatch.cpp
```diff
@@ -0,0 +1,19 @@
...
+time_t SkipInvalidTransactionsPatch::activationTimestamp;
+time_t SkipInvalidTransactionsPatch::lastBlockTimestamp;
+
+bool SkipInvalidTransactionsPatch::activateTimestampPassed() {
+    if ( activationTimestamp == 0 )
+        return false;
+    return lastBlockTimestamp >= activationTimestamp;
+}
+
+bool SkipInvalidTransactionsPatch::needToKeepTransaction( const Transaction& tr, bool excepted ) {
```
kladkogex: Please remove tr and also use underscore for function variables

libethereum/Client.cpp
```diff
@@ -654,7 +656,7 @@ size_t Client::syncTransactions(
     RevertableFSPatch::lastBlockTimestamp = blockChain().info().timestamp();
     StorageDestructionPatch::lastBlockTimestamp = blockChain().info().timestamp();
     POWCheckPatch::lastBlockTimestamp = blockChain().info().timestamp();
-
+    SkipInvalidTransactionsPatch::setLastBlock( blockChain().info() );
```
kladkogex: Please follow the logic of other patches  above - the code needs to be consistent

libskale/SkipInvalidTransactionsPatch.h
```diff
@@ -0,0 +1,43 @@
...
+// 2) eth_getTransactionReceipt (affects "transactionIndex" field)
+// 3) eth_getBlockTransactionCountByHash/Number
+// 4) eth_getTransactionByHash (invalid transactions are treated as never present)
+// 5) eth_getTransactionByBlockHash/NumberAndIndex
+// Transactions are removed from Transaction Queue as usually.
+
+class SkipInvalidTransactionsPatch : public SchainPatch {
+public:
+    static void init( const dev::eth::ChainParams& cp ) {
+        activationTimestamp = cp.sChain.skipInvalidTransactionsPatchTimestamp;
```
kladkogex: Please add setter and print logs as other patch

kladkogex (changes requested):
Please check the case if there is a single invalid transaction in the block.

kladkogex (changes requested):
Please check the case if there is a single invalid transaction in the block.

libethereum/Client.cpp
dimalit: It seems bad idea to stick just to timestamps - in general, patches can have different logic of activation. E.g. see AmsterdamFixPatch.

libskale/SkipInvalidTransactionsPatch.h
dimalit: It seems bad idea to just stick to timestamp-only approach. Different patches can have different logic of activation. E.g. see AmsterdamFixPatch

dimalit:
Case of single invalid transaction in block is checked in test/unittests/libethereum/SkaleHost.cpp

kladkogex (approved):

kladkogex (approved):

DmytroNazarenko (approved):

PR #1691, opened 2023-10-16
IS-869 Fix problem with new block removal

dimalit:
After GappedTransactionIndexCache is adding new block to itself - it checks cache overflow. This overflow check can lead to removal of the very same block that was just added. This PR fixes this


kladkogex (approved):

DmytroNazarenko (approved):

dimalit:
Need more bugfixiing, don't merge

PR #1695, opened 2023-10-17
Bug/skaled 1623 sigterm at exit

### LINKED ISSUE #1623: SIGETRM during shutdown forces shutdown
When skaled is shutting down by internal cause, it counts this event as though it was external SIGTERM. This leads to forced shutdown if real SIGTERM is received during shutdown procedure. Need not to count internal exit request as SIGTERM.

dimalit:
When skaled is shutting down by internal cause, it counts this event as though it was external SIGTERM. This leads to forced shutdown if real SIGTERM is received during shutdown procedure. Need not to count internal exit request as SIGTERM.

Test https://github.com/skalenetwork/skale-node-tests/blob/master/test_stop.py#L252


dimalit:
Probably, let's wait for "stable" - it will integrate changes in develop into v3.18.0 - and this PR would not be so bloated

libdevcore/Common.cpp
```diff
@@ -47,11 +47,18 @@ void ExitHandler::exitHandler( int s ) {
...
 void ExitHandler::exitHandler( int nSignalNo, ExitHandler::exit_code_t ec ) {
-    std::string strMessagePrefix = ExitHandler::shouldExit() ?
-                                       cc::error( "\nStop flag was already raised on. " ) +
-                                           cc::fatal( "WILL FORCE TERMINATE." ) +
-                                           cc::error( " Caught (second) signal. " ) :
-                                       cc::error( "\nCaught (first) signal. " );
+    std::string strMessagePrefix;
+    if ( nSignalNo > 0 ) {
+        strMessagePrefix = ( ExitHandler::shouldExit() && s_nStopSignal > 0 ) ?
+                               cc::error( "\nStop flag was already raised on. " ) +
```
kladkogex: Please remove color comments

libdevcore/Common.cpp
```diff
@@ -47,11 +47,18 @@ void ExitHandler::exitHandler( int s ) {
...
+                                   cc::error( " Caught (second) signal. " ) :
+                               cc::error( "\nCaught (first) signal. " );
+    } else {
+        strMessagePrefix = ExitHandler::shouldExit() ?
+                               cc::error( "\nInternal exit requested while already exiting. " ) :
+                               cc::error( "\nInternal exit initiated. " );
+    }
     std::cerr << strMessagePrefix << cc::error( skutils::signal::signal2str( nSignalNo ) )
               << "\n\n";
     std::cerr.flush();
```
kladkogex: Remove this line since cerr is always flushed on line end

libdevcore/Common.cpp
```diff
@@ -153,7 +160,8 @@ void ExitHandler::exitHandler( int nSignalNo, ExitHandler::exit_code_t ec ) {
 
     // nice exit here:
 
-    if ( ExitHandler::shouldExit() ) {
+    // TODO deduplicate with first if()
+    if ( ExitHandler::shouldExit() && s_nStopSignal > 0 && nSignalNo > 0 ) {
         std::cerr << ( "\n" + cc::fatal( "SIGNAL-HANDLER:" ) + " " +
                        cc::error( "Will force exit now..." ) + "\n\n" );
         _exit( 13 );
```
kladkogex, line 162: Why is 13 used here ? Is there a document in which error codes are defined?

kladkogex (changes requested):

DmytroNazarenko (dismissed):
Approved, next steps:
1. Need to add test to functional-tests
2. Fix colorization
4. Test build on devnet

dimalit:
UPD Currently, functional tests fail in load_js. But fail happens because of skaled JSON-RPC request refusal. As changes in this PR cannot affect JSON-RPC at all - most probably, issue is because of Debug build instead of Release build. It should be gone after merge and Release build.

DmytroNazarenko (approved):

PR #1698, opened 2023-10-18
IS-869 Fixed filtering condition in Eth.cpp

dimalit:
1. changed semantics of function that checks if there can be invalid transactions in block
2. added read-write lock

Testing: tested by Sasha1 on devnet


DmytroNazarenko (approved):

kladkogex (changes requested):
Please test the fix and specify how it was tested

kladkogex (approved):

PR #1713, opened 2023-10-31
Bug/1702 adopt precompiled oracle

### LINKED ISSUE #1702: adopt some of the precompiled contracts to process schain groups logic
need to adapt some of the precompiled smart contracts like `getConfigVariableUint256` to return the variables values depending on current block timestamp
```[tasklist]
### Tasks
- [ ] https://github.com/skalenetwork/skale-admin/issues/1000
```

How it works: https://github.com/skalenetwork/internal-support/blob/main/docs/issues-and-investigations/precompiledPatch.md

olehnikolaiev:
fixes #1702 

added two category of test in unit tests
the first one in ClientTest.cpp to test that precompiled contracts return different value depending on current block timestamp
the second one in PrecompiledTests.cpp to test that skaled properly parses historical data in config

tested manually on local machine as following:
- deploy oracle contract
- send oracle request and wait for response
- hard-code response to the attached contract. it will submit the response to `verifyOracleResponse` function
- send txn to the contract that do the `checkOracleResponse` call. see result
[test.txt](https://github.com/skalenetwork/skaled/files/14196821/test.txt)

to verify on qa network:
- deploy oracle contract
- stop one of the skaled containers
- send oracle request and wait for response  (you can use `oracle-demo` repo for it)
- hard-code response to the attached contract. it will submit the response to `verifyOracleResponse` function
- send txn to the contract that do the `checkOracleResponse` call. see result
- do several node rotations on the chain
- turn on the container that was stopped. wait till it catchup the txn with `checkOracleResponse` function and check its status.
- wait for 2 snapshots interval to verify that stateRoot is the same on all nodes


kladkogex (changes requested):
Please use a patch mechanism to switch the code changes on at a particular block id

Also please provided a detailed explaination in pull request description how the pull request was tested

kladkogex (changes requested):
please either add detailed comments how the fix was made (preferred)
Or describe details of the fix in pull request

libethereum/Precompiled.cpp
```diff
@@ -677,19 +679,21 @@ static const std::list< std::string > g_listReadableConfigParts{ "sealEngine",
     //"genesis.*"
     //"params.*",
 
-    "skaleConfig.nodeInfo.wallets.ima.commonBLSPublicKey*",
-    "skaleConfig.nodeInfo.wallets.ima.BLSPublicKey*",
+    // skaled-1702
```
kladkogex: Please remove the commented out section

According to the code guidelines we do not keep dead code

libethereum/Precompiled.cpp
```diff
@@ -677,19 +679,21 @@ static const std::list< std::string > g_listReadableConfigParts{ "sealEngine",
     //"genesis.*"
```
kladkogex: Please remove sealEngine parameter, since it is never used, and is always equal to ethHash

kladkogex: Please change stat_is_accessible_json_path fuction to use boost starts_with

libethereum/Precompiled.cpp
```diff
@@ -756,6 +760,27 @@ static dev::u256 stat_parse_u256_hex_or_dec( const std::string& strValue ) {
     return uValue;
 }
 
+static bool isCallToHistoricData( const std::string& callData ) {
+    // in C++ 20 there is string::starts_with, but we do not use C++ 20 yet
+    return boost::algorithm::starts_with( callData, "skaleConfig.sChain.nodes." );
+}
+
+static std::pair< std::string, unsigned > parseHistoricFieldReuqest( const std::string& callData ) {
```
kladkogex: Please improve so that exact matches are always used, so inccorect strings do not result in a match

libethereum/Precompiled.cpp
```diff
@@ -767,18 +792,37 @@ ETH_REGISTER_PRECOMPILED( getConfigVariableUint256 )( bytesConstRef _in ) {
...
+            } else {
+                throw std::runtime_error( "Incorrect config field" );
+            }
+        } else {
+            nlohmann::json joConfig = g_configAccesssor->getConfigJSON();
+            nlohmann::json joValue =
+                skutils::json_config_file_accessor::stat_extract_at_path( joConfig, rawName );
+            strValue = skutils::tools::trim_copy(
+                joValue.is_string() ? joValue.get< std::string >() : joValue.dump() );
+        }
```
kladkogex, line 832: Please change so that the else {} part is only used when the patch is enabled, since stat_extract_at_path is unsecure and complex. If it is needed, it needs to be rewritten in a clean and secure 
way using standard C++ and boost libraries

libethereum/Precompiled.cpp
```diff
@@ -807,11 +851,30 @@ ETH_REGISTER_PRECOMPILED( getConfigVariableAddress )( bytesConstRef _in ) {
...
+
         dev::u256 uValue( strValue.c_str() );
```
kladkogex: c_str() is not needed here since the constructor takes std::string

libethereum/Precompiled.cpp
```diff
@@ -677,19 +679,21 @@ static const std::list< std::string > g_listReadableConfigParts{ "sealEngine",
     //"genesis.*"
     //"params.*",
```
kladkogex: Please correct stat_parse_u256_hex_or_dec() function.

The constructor of u256 boost::multiprecision::number can take both decimal and hex string and autodetect. 

Therefore, this function can be simplified to just calling constructor of u256

libethereum/Precompiled.cpp
```diff
@@ -677,19 +679,21 @@ static const std::list< std::string > g_listReadableConfigParts{ "sealEngine",
...
```
kladkogex: Please make sure that the precompiled call only succeeds if the correct parameter (parameter name) is provided by the caller. In all other cases the call should revert

libethereum/Precompiled.cpp
```diff
@@ -677,19 +679,21 @@ static const std::list< std::string > g_listReadableConfigParts{ "sealEngine",
...
 
-    "skaleConfig.nodeInfo.wallets.ima.commonBLSPublicKey*",
-    "skaleConfig.nodeInfo.wallets.ima.BLSPublicKey*",
+    // skaled-1702
+    // remove these config field from public access for security reasons
```
kladkogex: Please remove all dead code in comments as per code guidelines.

libethereum/Precompiled.cpp
```diff
@@ -677,19 +679,21 @@ static const std::list< std::string > g_listReadableConfigParts{ "sealEngine",
...
+    //    "skaleConfig.nodeInfo.wallets.ima.commonBLSPublicKey*",
```
kladkogex: Function parseHistoricFieldReuqest should parse callData into 

Json::Value object (or into nlohmann json object)

And then verify that the object contains the required fields.

It should not use log level string arithmetic

kladkogex (dismissed):

dimalit (dismissed):

kladkogex (dismissed):

dimalit (dismissed):

DmytroNazarenko (approved):

dimalit (approved):

kladkogex (approved):

PR #1747, opened 2023-12-08
Bug/1744 handle custom errors outside evm

### LINKED ISSUE #1744: Handle custom errors thrown by evm
currently custom errors cannot be proccessed as revert reason, need to change it

olehnikolaiev:
fixes https://github.com/skalenetwork/skaled/issues/1744

added unittests at `test/unittests/libweb3jsonrpc/jsonrpc.cpp: revertReason` to verify handling different revert messages and custom errors

### DIFF
--- a/libdevcore/CommonData.cpp
+++ b/libdevcore/CommonData.cpp
@@ -24,7 +24,6 @@
 #include "CommonData.h"
 #include "Exceptions.h"
 
-#include <arpa/inet.h>
 #include <random>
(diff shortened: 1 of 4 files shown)


kladkogex (changes requested):
Please make it compatible to the way it is done in geth.

In particular, for error message one simply needs to return a byte array as is.

There is no need for any processing. 

Please look at geth code

PR #1757, opened 2023-12-15
IS-450 Reload rotation timestamp on each isExitTime request

badrogger:
### Problem 
The solution for https://github.com/skalenetwork/internal-support/issues/450 making skale-admin will write it to corresponding file directly instead of sending rpc call. The issue is that InstanceMonitor.cpp loads this file once during startup, which effectively says that timestamp won't be passed to skaled at all. 

### Changes
`isExitTime` will check the content of rotation.txt file and compare it to current block timestamp instead on relaying on `m_finishTimestamp` field.

### Performance 
Since `rotation.txt` file is only 4 bytes and reading is happening in background performance should not be affected. 
### Unit tests
Modified tests in `InstanceMonitorTest.cpp`.

### Testing 
Tested by running rotation procedure on local network.

### DIFF
--- a/libethereum/InstanceMonitor.cpp
+++ b/libethereum/InstanceMonitor.cpp
@@ -46,23 +46,20 @@ void InstanceMonitor::initRotationParams( uint64_t _finishTimestamp ) {
     std::ofstream rotationInfoFile( m_rotationInfoFilePath.string() );
     rotationInfoFile << rotationJson;
 
-    m_finishTimestamp = _finishTimestamp;
-    LOG( m_logger ) << "Set rotation time to " << m_finishTimestamp;
+    LOG( m_logger ) << "Set rotation time to " << _finishTimestamp;
 }
 
 bool InstanceMonitor::isTimeToRotate( uint64_t _finishTimestamp ) {
     if ( !fs::exists( m_rotationInfoFilePath ) ) {
         return false;
     }
-    return m_finishTimestamp <= _finishTimestamp;
+    return getRotationTimestamp() <= _finishTimestamp;
 }
 
-void InstanceMonitor::restoreRotationParams() {
-    if ( fs::exists( m_rotationInfoFilePath ) ) {
-        std::ifstream rotationInfoFile( m_rotationInfoFilePath.string() );
-        auto rotationJson = nlohmann::json::parse( rotationInfoFile );
-        m_finishTimestamp = rotationJson["timestamp"].get< uint64_t >();
-    }
+uint64_t InstanceMonitor::getRotationTimestamp() {
+    std::ifstream rotationInfoFile( m_rotationInfoFilePath.string() );
+    auto rotationJson = nlohmann::json::parse( rotationInfoFile );
+    return rotationJson["timestamp"].get< uint64_t >();
 }
 
 void InstanceMonitor::reportExitTimeReached( bool _reached ) {
--- a/libethereum/InstanceMonitor.cpp
+++ b/libethereum/InstanceMonitor.cpp
@@ -53,10 +53,10 @@ bool InstanceMonitor::isTimeToRotate( uint64_t _finishTimestamp ) {
     if ( !fs::exists( m_rotationInfoFilePath ) ) {
         return false;
     }
-    return getRotationTimestamp() <= _finishTimestamp;
+    return rotationTimestamp() <= _finishTimestamp;
 }
 
-uint64_t InstanceMonitor::getRotationTimestamp() {
+uint64_t InstanceMonitor::rotationTimestamp() const {
     std::ifstream rotationInfoFile( m_rotationInfoFilePath.string() );
     auto rotationJson = nlohmann::json::parse( rotationInfoFile );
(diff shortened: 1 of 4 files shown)


libethereum/InstanceMonitor.cpp
```diff
@@ -46,23 +46,20 @@ void InstanceMonitor::initRotationParams( uint64_t _finishTimestamp ) {
...
-    m_finishTimestamp = _finishTimestamp;
-    LOG( m_logger ) << "Set rotation time to " << m_finishTimestamp;
+    LOG( m_logger ) << "Set rotation time to " << _finishTimestamp;
 }
 
 bool InstanceMonitor::isTimeToRotate( uint64_t _finishTimestamp ) {
     if ( !fs::exists( m_rotationInfoFilePath ) ) {
         return false;
     }
-    return m_finishTimestamp <= _finishTimestamp;
```
DmytroNazarenko, line 57: It's better to write information about rotation into logs

libethereum/InstanceMonitor.cpp
```diff
@@ -46,23 +46,20 @@ void InstanceMonitor::initRotationParams( uint64_t _finishTimestamp ) {
     std::ofstream rotationInfoFile( m_rotationInfoFilePath.string() );
     rotationInfoFile << rotationJson;
 
-    m_finishTimestamp = _finishTimestamp;
-    LOG( m_logger ) << "Set rotation time to " << m_finishTimestamp;
+    LOG( m_logger ) << "Set rotation time to " << _finishTimestamp;
 }
 
 bool InstanceMonitor::isTimeToRotate( uint64_t _finishTimestamp ) {
```
DmytroNazarenko: Better to rename the parameter to `currentTimestamp`/`blockTimestamp` or sth like this

DmytroNazarenko (changes requested):

libethereum/InstanceMonitor.cpp
```diff
@@ -46,23 +46,22 @@ void InstanceMonitor::initRotationParams( uint64_t _finishTimestamp ) {
...
-void InstanceMonitor::restoreRotationParams() {
-    if ( fs::exists( m_rotationInfoFilePath ) ) {
-        std::ifstream rotationInfoFile( m_rotationInfoFilePath.string() );
-        auto rotationJson = nlohmann::json::parse( rotationInfoFile );
-        m_finishTimestamp = rotationJson["timestamp"].get< uint64_t >();
-    }
+uint64_t InstanceMonitor::rotationTimestamp() const {
+    std::ifstream rotationInfoFile( m_rotationInfoFilePath.string() );
+    auto rotationJson = nlohmann::json::parse( rotationInfoFile );
+    auto timestamp = rotationJson["timestamp"].get< uint64_t >();
```
kladkogex: Add try catch to handle the case of corrupt file

kladkogex (commented):
Change .txt to .json

libethereum/InstanceMonitor.cpp
```diff
@@ -46,23 +46,22 @@ void InstanceMonitor::initRotationParams( uint64_t _finishTimestamp ) {
     std::ofstream rotationInfoFile( m_rotationInfoFilePath.string() );
     rotationInfoFile << rotationJson;
 
-    m_finishTimestamp = _finishTimestamp;
-    LOG( m_logger ) << "Set rotation time to " << m_finishTimestamp;
+    LOG( m_logger ) << "Set rotation time to " << _finishTimestamp;
```
kladkogex: Add try catch

libethereum/InstanceMonitor.h
```diff
@@ -61,5 +56,5 @@ class InstanceMonitor {
     void reportExitTimeReached( bool _reached );
 
 private:
-    dev::Logger m_logger{ createLogger( dev::VerbosityInfo, "instance-monitor" ) };
+    mutable dev::Logger m_logger{ createLogger( dev::VerbosityInfo, "instance-monitor" ) };
```
kladkogex: mutable is not needed here

dimalit (dismissed):

kladkogex (dismissed):

dimalit (dismissed):

libethereum/InstanceMonitor.h
```diff
@@ -36,30 +36,40 @@ class InstanceMonitor {
...
+            what_str = "File " + path.string() + " is malformed or missing";
+        }
+        virtual const char* what() const noexcept override { return what_str.c_str(); }
+    };
+
+
 private:
-    dev::Logger m_logger{ createLogger( dev::VerbosityInfo, "instance-monitor" ) };
+    mutable dev::Logger m_info_logger{ createLogger( dev::VerbosityInfo, "instance-monitor" ) };
+    mutable dev::Logger m_error_logger{ createLogger( dev::VerbosityError, "instance-monitor" ) };
```
DmytroNazarenko: Change to camelCase

DmytroNazarenko (changes requested):

DmytroNazarenko (approved):

olehnikolaiev (approved):

PR #1768, opened 2023-12-19
Bug/skaled 1745 invalid fork in pow gas

### LINKED ISSUE #1745: Invalid gas computation/fork block in checkOutEternalGas
checkOutEternalGas uses Constantinople:
`if ( externalGas >= baseGasRequired( ConstantinopleSchedule ) )` (libethereum/Transaction.cpp:193)

This can lead to rejection of PoW transactions with enough, but not that many gas.

For example, transaction with 50kb of data would actually require 800k of gas, but PoW mechanism will reject it if supplied less than 3.6M gas.
See also #1600 

This will need to be done as a patch so the state roots do not diverge

dimalit:
Changes:
1 `Transaction::checkOutExternalGas()` now operates on current Ethereum fork, instead of Constantinople (base gas for 1 byte of data changed from 68 gas/b to 16 gas/b)
2 All changes are activated after `correctForkInPowPatchTimestamp`. This is true both for external gas check and for eth_estimateGas call.
3 `Transaction::checkOutExternalGas()` is called one additional time after consensus (needed to be sure that patch activates smoothly). This make possible little decrease of block processing performance (but unlikely).

Testing:
JsonRpcSuite/simplePoWTransaction tests that PoW transaction with incorrect amount of gas fails before patch timestamp and works after it

### DIFF
--- a/libethereum/Transaction.cpp
+++ b/libethereum/Transaction.cpp
@@ -198,9 +198,11 @@ void Transaction::checkOutExternalGas( const ChainParams& _cp, uint64_t _bn ) {
             scheduleForUse = _cp.scheduleForBlockNumber( _bn );
 
         // !! never call checkOutExternalGas with non-last block!
-        if ( _bn != CorrectForkInPowPatch::getLastBlockNumber() )
+        if ( _bn != CorrectForkInPowPatch::getLastBlockNumber() ) {
+            ctrace << _bn << " != " << CorrectForkInPowPatch::getLastBlockNumber();
             BOOST_THROW_EXCEPTION( std::runtime_error(
                 "Internal error: checkOutExternalGas() has invalid block number" ) );
+        }
 
         if ( externalGas >= baseGasRequired( scheduleForUse ) )
             m_externalGas = externalGas;
--- a/libethereum/Transaction.cpp
+++ b/libethereum/Transaction.cpp
@@ -197,7 +197,7 @@ void Transaction::checkOutExternalGas( const ChainParams& _cp, uint64_t _bn ) {
         if ( CorrectForkInPowPatch::isEnabled() )
             scheduleForUse = _cp.scheduleForBlockNumber( _bn );
 
-        // !! never call checkOutExternalGas with non-last block!
+        // never call checkOutExternalGas with non-last block
         if ( _bn != CorrectForkInPowPatch::getLastBlockNumber() ) {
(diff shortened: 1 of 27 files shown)


libethereum/Transaction.cpp
```diff
@@ -180,19 +181,32 @@ u256 Transaction::gasPrice() const {
...
         if ( externalGas > 0 )
             ctrace << "Mined gas: " << externalGas << endl;
-        if ( externalGas >= baseGasRequired( ConstantinopleSchedule ) ) {
-            m_externalGas = externalGas;
+
+        EVMSchedule scheduleForUse = ConstantinopleSchedule;
+        if ( CorrectForkInPowPatch::isEnabled() )
+            scheduleForUse = _cp.scheduleForBlockNumber( _bn );
+
+        // !! never call checkOutExternalGas with non-last block!
```
DmytroNazarenko: Please, remove exclamation marks from code

libethereum/Transaction.cpp
```diff
@@ -180,19 +181,32 @@ u256 Transaction::gasPrice() const {
...
+        if ( _bn != CorrectForkInPowPatch::getLastBlockNumber() ) {
+            ctrace << _bn << " != " << CorrectForkInPowPatch::getLastBlockNumber();
```
DmytroNazarenko, line 202: Please use `LOG( m_traceLogger )` instead

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -225,7 +225,7 @@ bytes ImportTest::executeTest( bool _isFilling ) {
             continue;
 
         for ( auto& tr : m_transactions ) {
-            tr.transaction.checkOutExternalGas( 100 );
+//            tr.transaction.checkOutExternalGas( 100 );
```
DmytroNazarenko: Please, remove commented lines of code (or uncomment)

DmytroNazarenko (changes requested):

libethereum/Transaction.cpp
dimalit: there is no m_traceLogger in Transaction

DmytroNazarenko (commented):
1. Please, add patch timestamp to estimateGas
2. Update description with details about new patch timestamp

olehnikolaiev (dismissed):

DmytroNazarenko (dismissed):

DmytroNazarenko (approved):

olehnikolaiev (approved):

PR #1868, opened 2024-04-10
Enhancement/skaled 1583 patch architecture plus 17.2

### LINKED ISSUE #1583: Patch timestamp architecture
Need to create some smart solution for backward compatibility: when schain is upgraded to new skaled version, this version should at first work as previous one - and after all have been upgraded - start to work in a new way. Currently, we use ContractStorageZeroValuePatch, ContractStorageZeroValuePatch etc. Probably we need "magic transaction" solution.
```[tasklist]
### Tasks
- [ ] https://github.com/skalenetwork/skaled/issues/1846
```

### LINKED ISSUE #1851: eth_call ignores patches
When we use block timestamps to detect patches activation, it works in transactions, but is ignored in eth_call.
Problem is relevant only to new patch architecture (in old architecture it will use latest patches even on old blocks).

Test:
```
// SPDX-License-Identifier: GPL-3.0

pragma solidity >=0.8.2;

contract Push0Test {
    function f() public pure returns (string memory){
        return "success";
    }
}
```
Build it for Shanghai fork, deploy on a chain with PushZeroPatch, call f().
Expected: "success"
Actual: empty output + "bad instruction" exception

dimalit:
fixes #1583 

## Description

1. "Forks" are now activated by block timestamp, rather than number.
2. There are DEFINE_XXX_PATCH macros for defining patches.
3. Timestamps for patches are automatically read from config.
4. Existing patches use these macros.

## Tests

Tests for patches:
 - ContractStorageLimitPatch: storage_limit_reverted
 - ContractStorageZeroValuePatch: tested manually, used in test storage_limit_predeployed
 - RevertableFSPatch: FilestorageCacheSuite)
 - StorageDestructionPatch: tested manually with `JsonRpcSuite/simple_contract`
 - POWCheckPatch: powTxnGasLimit
 - PushZeroPatch: LegacyVMSuite/InstructionSuite
 - SkipInvalidTransactionsPatch: SkaleHostSuite
 - PrecompiledConfigPatch: PrecompiledTests/getConfigVariable
 - CorrectForkInPowPatch: simplePoWTransaction
 - VerifyDaSigsPatch - need to check manually!

## Performance Impact

There can be significant performance impact, as patch activations are evaluated at each block or sometimes even every transaction. Quick testing with stability tests (skale-ci-intergration_tests) showed no significant difference between 3.19.0-develop-latest and 3.19.0-develop.3

## Details about changes in code

1. A lot of functions that needed only `blockNumber` parameter now need also `committedBlockTimestamp`.
2. Removed 'Frontier' fork (it was never used), as well as 'SkaleNNNForkBlock`s for same reson.
3. `EVMSchedule` can now be constructed from several patches, and not only taken from pre-defined list as before (`ChainOperationParams::evmSchedule`)
4. Some functionality moved from `SealEngine` to `ChainOperationParams`, and `SealEngine` is now unneeded in some places, where `ChainOperationParams` suffice - for simplification of architecture. Also it's planned to cut out SealEngine completely.
5. `PushZeroPatch` uses boolean field in `EVMSchedule`, and generates new EVMSchedule similar to 'fork blocks'
6. In `libethereum/SchainPatch.h` there are new macros for creation of patches.
7. Almost all old patches now use these macros (see SchainPatch.h)
8. Latest block timestamp is updated every block for all patches at once - in `SchainPatch` class.


kladkogex (dismissed):

olehnikolaiev (dismissed):

libethereum/ClientBase.cpp
```diff
@@ -91,12 +91,13 @@ std::pair< bool, ExecutionResult > ClientBase::estimateGasStep( int64_t _gas, Bl
     t.forceSender( _from );
     t.forceChainId( chainId() );
     t.ignoreExternalGas();
-    EnvInfo const env( _latestBlock.info(), bc().lastBlockHashes(), 0, _gas );
+    EnvInfo const env( _pendingBlock.info(), bc().lastBlockHashes(),
+        _pendingBlock.previousInfo().timestamp(), 0, _gas );
```
DmytroNazarenko, line 95: Mb `_pendingBlock.previousInfo().timestamp()` -> `_latestBlock.info().timestamp()` would be better?

libethereum/ClientBase.cpp
```diff
@@ -116,15 +117,17 @@ std::pair< u256, ExecutionResult > ClientBase::estimateGas( Address const& _from
...
-                Transaction::baseGasRequired( !_dest, &_data, EVMSchedule() );
-
-        Block bk = latestBlock();
-        if ( upperBound > bk.info().gasLimit() ) {
-            upperBound = bk.info().gasLimit().convert_to< int64_t >();
+        int64_t lowerBound = CorrectForkInPowPatch::isEnabledInWorkingBlock() ?
+                                 Transaction::baseGasRequired( !_dest, &_data,
+                                     bc().sealEngine()->chainParams().makeEvmSchedule(
+                                         bc().info().timestamp(), bc().number() ) ) :
+                                 Transaction::baseGasRequired( !_dest, &_data, EVMSchedule() );
```
DmytroNazarenko: if/else would be more readable here

libethereum/SkaleHost.cpp
```diff
@@ -447,8 +445,11 @@ ConsensusExtFace::transactions_vector SkaleHost::pendingTransactions(
                 try {
                     bool isMtmEnabled = m_client.chainParams().sChain.multiTransactionMode;
                     Executive::verifyTransaction( tx,
+                        static_cast< const Interface& >( m_client )
+                            .blockInfo( LatestBlock )
+                            .timestamp(),
```
DmytroNazarenko: This call could be retrieved and executed before loop

libethereum/SkaleHost.cpp
```diff
@@ -653,7 +654,8 @@ void SkaleHost::createBlock( const ConsensusExtFace::transactions_vector& _appro
             // TODO clear occasionally this cache?!
             if ( m_m_transaction_cache.find( sha.asArray() ) != m_m_transaction_cache.cend() ) {
                 Transaction t = m_m_transaction_cache.at( sha.asArray() );
-                t.checkOutExternalGas( m_client.chainParams(), m_client.number(), true );
+                t.checkOutExternalGas( m_client.chainParams(),
+                    m_client.blockInfo( LatestBlock ).timestamp(), m_client.number(), true );
```
DmytroNazarenko: `m_client.blockInfo( LatestBlock ).timestamp()` can be executed before loop

DmytroNazarenko (changes requested):

libethereum/ClientBase.cpp
dimalit: same

libethereum/ClientBase.cpp
dimalit: :+1:

libethereum/SkaleHost.cpp
dimalit: :ok_hand:

libethereum/SkaleHost.cpp
dimalit: :+1:

kladkogex (commented):
Please remove exclamation marks and dead code

kladkogex (approved):

DmytroNazarenko (approved):

PR #1969, opened 2024-08-23
SKALED-1900 release build with separate debug info

### LINKED ISSUE #1900: Remove debug library builds from release
It looks like the release build includes some libraries in debug  (ZeroMQ) mode.

This needs to be fixed since it may affects performance

We need a check at the end of the build that the executable does not have debug symbols

image

dimalit:
1. All dependencies, are built with `-O3` and without `-g`.
2. `skaled` is built with `-O3` but WITH `-g`.
3. `-rdynamic` removed.
4. `skaled` binary is stripped, but it's copy wit debug info is kept.

CUSTOM BUILD Now contains option to build maximally optimized "Release" configuration

Consequences:
 1. Stack trace printed by skaled on crash will not contain any function names.
 2. gdb will be able to use `skaled-debug` file to load debug info for original skaled
 3. `addr2line` will be able to convert addresses from skaled's printed stack trace into source code locations using `skaled-debug` binary

Testing:
1
```
objdump -h skaled | grep debug
nm -a skaled
```
2 Execute `strip` command on `skaled` and see that it's size didn't change
3 `ojdump -h <file> | grep debug` on all project libraries shows no result too.

`skaled` binary size `29575576`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -118,7 +118,6 @@ jobs:
           make skaled -j$(nproc)
           #echo "Ensure release mode skaled does not have any debug markers"
           cp skaled/skaled skaled/skaled-debug
-          strip skaled/skaled
           cd ..
       - name: Configure historic state build
         run: |
@@ -141,7 +140,6 @@ jobs:
           make skaled -j$(nproc)
           #echo "Ensure release mode skaled does not have any debug markers"
           cp skaled/skaled skaled/skaled-debug
-          strip skaled/skaled
           cd ..
       - name: Build and publish container
         run: |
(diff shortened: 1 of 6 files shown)


kladkogex (dismissed):

.github/workflows/publish.yml
```diff
@@ -28,6 +28,7 @@ jobs:
         run: |
           sudo apt-get -y remove libzmq* || true
           sudo apt-get -y install software-properties-common gcc-9 g++-9 || true
+          sudo apt-get -y install liblz4-dev || true
```
olehnikolaiev: maybe we can add this library to our deps script instead of installing in through apt-get? seems like it is pretty easy to install https://github.com/lz4/lz4

kladkogex (approved):

olehnikolaiev (approved):

PR #1999, opened 2024-09-19
IS 1074 fix externalGas calculation

olehnikolaiev:
fixes https://github.com/skalenetwork/internal-support/issues/1074

fixed `externalGas` calculation, added patch `externalGasPatch` for it

added unit test that send txns from a suspicious accounts before and after patch activation
ran load tests with POW transactions on devnet for a week - bug hasn't been reproduced

no performance impact

### DIFF
--- a/libethereum/Transaction.cpp
+++ b/libethereum/Transaction.cpp
@@ -200,14 +200,18 @@ void Transaction::checkOutExternalGas( const ChainParams& _cp, time_t _committed
     u256 const& difficulty = _cp.externalGasDifficulty;
     assert( difficulty > 0 );
     if ( ( _force || !m_externalGasIsChecked ) && !isInvalid() ) {
-        // reset externalGas value
-        // we may face patch activation after txn was added to the queue but before it was executed
-        // therefore we need to recalculate externalGas
-        if ( m_externalGasIsChecked && hasExternalGas() )
-            m_externalGas = 0;
-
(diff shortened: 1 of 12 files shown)


libethereum/Transaction.cpp
```diff
@@ -200,7 +200,14 @@ void Transaction::checkOutExternalGas( const ChainParams& _cp, time_t _committed
     u256 const& difficulty = _cp.externalGasDifficulty;
     assert( difficulty > 0 );
     if ( ( _force || !m_externalGasIsChecked ) && !isInvalid() ) {
-        h256 hash = dev::sha3( sender().ref() ) ^ dev::sha3( nonce() ) ^ dev::sha3( gasPrice() );
+        // reset externalGas value
+        // we may face patch activation after txn was added to the queue but before it was executed
+        // therefore we need to recalculate externalGas
+        if ( m_externalGasIsChecked && hasExternalGas() )
```
dimalit: Why not just:
```
m_externalGasIsChecked = false;
m_externalGas.reset()'
```
without `if`?

dimalit (changes requested):
see comment

kladkogex (dismissed):

DmytroNazarenko (approved):

dimalit (approved):

kladkogex (approved):

PR #2072, opened 2025-01-08
Feature/1900 remove debug symbols release

### LINKED ISSUE #1900: Remove debug library builds from release

olehnikolaiev:
fixes #1900 

1. All dependencies, are built with -O3 and without -g
2. skaled is built with -O3 but WITH -g.
3. -rdynamic removed.
4. skaled binary is stripped, but it's copy wit debug info is kept.
5. asserts are removed from `RelWithDebInfo` build

Consequences:
1. Stack trace printed by skaled on crash will not contain any function names.
2. gdb will be able to use skaled-debug file to load debug info for original skaled
3. addr2line will be able to convert addresses from skaled's printed stack trace into source code locations using skaled-debug binary

Testing:
1
```
objdump -h skaled | grep debug
nm -a skaled
```

2. Execute strip command on skaled and see that it's size didn't change
3. ojdump -h <file> | grep debug on all project libraries shows no result too.

skaled binary size 29575576


.github/workflows/setup-build-publish.yml
```diff
@@ -150,6 +150,13 @@ jobs:
           export CMAKE_BUILD_TYPE=$BUILD_TYPE
           cd build
           make skaled -j$(nproc)
+          if [[ "$BUILD_TYPE" = "Release" ]]; then
+              debug_wc=$(objdump -h skaled/skaled | grep -i debug | wc -l)
+              sym_wc=$(readelf -s skaled/skaled | wc -l)
+              if (( debug_wc != 0 || sym_wc > 10000 )); then
+                  exit 1
+              fi
+          fi
```
DmytroNazarenko, line 176: `custom_build.yml` needs to be modified to support Release option

.github/workflows/publish.yml
DmytroNazarenko, line 1: Propose to use `setup-build-publish.yml` workflow underneath with Release parameter

DmytroNazarenko (changes requested):

DmytroNazarenko (dismissed):

DmytroNazarenko (approved):

DmytroNazarenko (approved):

kladkogex (approved):

PR #2075, opened 2025-01-13
1545 second half

### LINKED ISSUE #1545: Cleanup state locking and make JSON RPC calls like eth_call  faster

kladkogex:
## Description

1. Rewrote Unddos and stats implementation with a way  simpler one since the previous one had lots of bugs and deadlocks.
The new Unddos impelementation does not use any locks and can not cause a deadlock.

3. Made sure that tracing calls are only enabled when  tracing is enabled, previously tracing was affecting performance even if disabled.

4. When forming a block proposal from the pending queue, If two transactions have the same gas price and height related to the current nonce, we currently give priority to the older transaction. Previously the behavior was undefined, so old transactions could stay in the pending queue for a long time.

## Tests

There are now performance tests for all major read calls such as eth_call and eth_getBlock, as well as performance tests for transaction submission

The performance drastically improved.


libethereum/BlockChain.cpp
```diff
@@ -950,8 +950,8 @@ ImportRoute BlockChain::insertBlockAndExtras( VerifiedBlockRef const& _block,
     }
     pieceUsageBytes += writeSize;
 
-    LOG( m_loggerInfo ) << "Block " << tbi.number() << " DB usage is " << writeSize
-                        << ". Piece DB usage is " << pieceUsageBytes << " bytes";
+    LOG( m_loggerDetail ) << "BLOCK_DB+" << writeSize << ". Piece DB usage is " << pieceUsageBytes
```
olehnikolaiev, line 953: this information may be used by qa team or monitoring service. we need to make sure that moving this log to a different log level doesn't break anything

skaled/main.cpp
```diff
@@ -2454,18 +2456,30 @@ int main( int argc, char** argv ) try {
             // unddos
             if ( joConfig.count( "unddos" ) > 0 ) {
                 nlohmann::json joUnDdosSettings = joConfig["unddos"];
+                clog( VerbosityInfo, "main" ) << "DDOS config :" << joConfig["unddos"].dump();
```
olehnikolaiev: should we keep this log? it dumps the entire ddos config

libskale/httpserveroverride.cpp
```diff
@@ -270,240 +251,58 @@ bool checkParamsIsObject( const char* strMethodName, const rapidjson::Document&
...
+    // note that no lock is needed here since the map is read only after being populated at init
+    // and we use atomic ints
+
+
+    for ( auto&& iterator : dev::rpc::SkaleStats::statsCounters ) {
+        auto strSubsystemAndMethodName = iterator.first;
+        if ( boost::algorithm::starts_with( strSubsystemAndMethodName, _subSystem ) ) {
+            nlohmann::json joMethod = nlohmann::json::object();
+            auto methodName = strSubsystemAndMethodName.substr( _subSystem.size() );
+            LDB_CHECK( methodName.find( ':' ) == string::npos );
```
olehnikolaiev, line 267: lets not use `LDB_CHECK` here cause it throws `dev::db::DatabaseError()` on error

libweb3jsonrpc/SkaleStats.cpp
```diff
@@ -542,312 +479,44 @@ Json::Value SkaleStats::skale_imaInfo() {
...
-        if ( joSkaleConfig_nodeInfo.count( "no-ima-signing" ) == 0 )
-            throw std::runtime_error(
-                "error in config.json file, cannot find "
-                "\"skaleConfig\"/\"nodeInfo\"/\"no-ima-signing\"" );
-        const nlohmann::json& joSkaleConfig_nodeInfo_isEnabled =
-            joSkaleConfig_nodeInfo["no-ima-signing"];
-        isEnabled = joSkaleConfig_nodeInfo_isEnabled.get< bool >() ? false : true;
-    } catch ( ... ) {
+void SkaleStats::initStatsCounters() {
+    if ( !statsCounters.empty() > 0 ) {
```
olehnikolaiev, line 484: it can be simplified to `!statsCounters.empty()`

libskutils/src/ws.cpp
```diff
@@ -3342,9 +2817,7 @@ std::string peer::getShortTypeDescrition( bool isColored /*= false*/ ) const {
     }
     return s;
 }
-std::string peer::getShortPeerDescription(
-    bool isColored /*= false*/, bool isLifetime /*= true*/, bool isTrafficStats /*= true*/ ) const {
-    // lock_type lock( ref_mtx() );
+std::string peer::getShortPeerDescription( bool isColored /*= false*/, bool /*= true*/ ) const {
```
olehnikolaiev, line 2820: why dont we remove all parameters for this method as they are not used anyway?

test/unittests/libskutils/test_skutils_unddos.cpp
```diff
@@ -1,64 +1,80 @@
...
+        ++ttmNow;
+        BOOST_REQUIRE( unddos.register_call_from_origin( "11.11.11.11", ttmNow ) ==
+                       skutils::unddos::e_high_load_detection_result_t::ehldr_no_error );
     }
-    BOOST_REQUIRE( unddos.register_call_from_origin( "11.11.11.11", ttmNow ) != skutils::unddos::e_high_load_detection_result_t::ehldr_no_error );
+    BOOST_REQUIRE( unddos.register_call_from_origin( "11.11.11.11", ttmNow ) !=
+                   skutils::unddos::e_high_load_detection_result_t::ehldr_no_error );
 }
 
-BOOST_AUTO_TEST_CASE( ws_conn_counting ) {
```
olehnikolaiev, line 50: why was this test removed?

libskutils/src/http_pg.cpp
```diff
@@ -17,101 +17,99 @@
...
         return;
-    std::cout << s;
-}
-
-void pg_log( const std::string& s ) {
-    if ( s.empty() )
+    if ( _s.empty() )
         return;
-    return pg_log( s.c_str() );
+    std::cerr << _s << std::endl;
```
olehnikolaiev, line 52: why this log is written to cerr not cout or boost log?

libskale/httpserveroverride.cpp
```diff
@@ -2546,20 +1975,13 @@ bool SkaleServerOverride::implStartListening(  // web socket
...
-                cc::info( strAddr ) + cc::success( " and port " ) + cc::c( nPort ) +
-                cc::debug( "..." ) );
+        cnote << "Started server";
+
         return true;
     } catch ( const std::exception& ex ) {
-        logTraceServerEvent( false, ipVer, bIsSSL ? "WSS" : "WS", pSrv->serverIndex(), esm,
-            cc::fatal( "FAILED" ) + cc::error( " to start " ) + cc::warn( bIsSSL ? "WSS" : "WS" ) +
-                cc::error( " server: " ) + cc::warn( ex.what() ) );
+        cerror << "FAILED to start server: " << ex.what();
```
olehnikolaiev, line 1982: does the exception caught here has any information regarding what server at which port failed to start? if no, it will be great to have it as a part of the error message

libweb3jsonrpc/Debug.cpp
```diff
@@ -68,6 +68,9 @@ void Debug::debug_forceBroadcast( const string& _transactionHash ) {
 }
 
 string Debug::debug_interfaceCall( const string& _arg ) {
+    if ( !m_debugInterface ) {
+        BOOST_THROW_EXCEPTION( jsonrpc::JsonRpcException( "Debug interface disabled" ) );
+    }
```
DmytroNazarenko, line 73: What is the purpose of this check?

libweb3jsonrpc/Eth.cpp
```diff
@@ -138,19 +138,11 @@ string Eth::eth_coinbase() {
...
+    return toJS( 0 );
 }
 
 bool Eth::eth_mining() {
-    try {
-        return asEthashClient( client() )->isMining();
-    } catch ( InvalidSealEngine& ) {
-        BOOST_THROW_EXCEPTION( JsonRpcException( Errors::ERROR_RPC_INVALID_PARAMS ) );
-    }
+    return false;
```
DmytroNazarenko, line 145: What is the purpose of these replacements?

libweb3jsonrpc/SkaleStats.cpp
```diff
@@ -14,7 +14,7 @@
     GNU General Public License for more details.
 
     You should have received a copy of the GNU General Public License
-    along with skaled.  If not, see <http://www.gnu.org/licenses/>.
+    along with skaled.  If not, see <http:www.gnu.org/licenses/>.
```
DmytroNazarenko, line 17: typo

skaled/main.cpp
```diff
@@ -1994,9 +1994,11 @@ int main( int argc, char** argv ) try {
         auto pAdminEthFace = bEnabledAPIs_admin ? new rpc::AdminEth( *g_client, *gasPricer.get(),
                                                       keyManager, *sessionManager.get() ) :
                                                   nullptr;
+
         auto pDebugFace = bEnabledAPIs_debug ?
                               new rpc::Debug( *g_client, &debugInterface, argv_string ) :
                               nullptr;
+        SkaleDebugInterface::g_isEnabled = bEnabledAPIs_debug;
```
DmytroNazarenko, line 2001: What is the purpose of g_isEnabled?

test/unittests/libweb3jsonrpc/jsonrpc.cpp
```diff
@@ -3831,6 +3831,7 @@ BOOST_AUTO_TEST_CASE( etherbase_generation2 ) {
 
     // mine block without transactions
     dev::eth::simulateMining( *( fixture.client ), 1 );
+    sleep(3);
```
DmytroNazarenko, line 3904: Why this line is needed?

libethereum/BlockChain.cpp
kladkogex: I will ask

libweb3jsonrpc/Debug.cpp
kladkogex: We have a policy in code guidelines to check everything for null
before using it

If bad things for whatever reason happen this will cause exception and not a hard crash

libweb3jsonrpc/Eth.cpp
kladkogex: This is to make response fast since we do not support mining

skaled/main.cpp
kladkogex: This is a boolean variable that show that the debug interface is enabled.
I use it to enable tracing if debug is enabled. Before tracing was always enabled that slowed down things.

skaled/main.cpp
kladkogex: Lets deleted it, we do not need it

test/unittests/libskutils/test_skutils_unddos.cpp
kladkogex: Because I removed websocket connection counting ddos.  Now ddos is just based on requests  couinting, not on the number of websocket connections.

In the future we will move to proxygen for web sockets, our current websocket implementation is not perfect.

libethereum/BlockChain.cpp
kladkogex: The QA team and Alex say it is ok to move it to m_loggerDetail

DmytroNazarenko (approved):

olehnikolaiev (approved):

PR #2107, opened 2025-02-21
Fix state root mismatch error on 4.0.0

### LINKED ISSUE #2085: Fix state root mismatch in 4.0
Since we optimized the way partial receipts are handled in 2.0, looks like this caused state root mismatch.

Fix  the mismatch,  also add a test for it.

badrogger:
## Description
Keep the old way of saving partial receipts before `ClearPartialReceiptsPatchTimestamp`

## Tests
- Added unit tests
- Tested on local network by launching archive node with `syncFromCatchup=True` 

## Performance Impact
- Additional commit to DB added per block. No performance changes expected.

### DIFF
--- a/libethereum/Block.cpp
+++ b/libethereum/Block.cpp
@@ -595,6 +595,9 @@ tuple< TransactionReceipts, unsigned > Block::syncEveryone( BlockChain const& _b
          receiptsOfCommited.size() > 0 ) {
         // Saving partial receipts old way to be compatible with < 4.0 version
         m_state.safeCommitLegacyPartialTransactionReceipts( receiptsOfCommited );
+    } else {
+        // Making sure the old record was removed from the state db
+        m_state.safeRemoveLegacyPartialTransactionReceipts();
     }
 
     return make_tuple( receipts, receipts.size() - countBad );
--- a/libethereum/Block.cpp
+++ b/libethereum/Block.cpp
@@ -483,6 +483,8 @@ tuple< TransactionReceipts, unsigned > Block::syncEveryone( BlockChain const& _b
         m_receipts = m_state.safePartialTransactionReceipts( info().number() );
     TransactionReceipts receipts = m_receipts;
 
+    TransactionReceipts receiptsOfCommited;
+
     unsigned countBad = 0;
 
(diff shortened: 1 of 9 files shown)


libskale/State.cpp
```diff
@@ -321,6 +321,39 @@ void State::safeRemoveAllPartialTransactionReceipts() {
...
+
+
+dev::eth::TransactionReceipts State::safeLegacyPartialTransactionReceipts() {
+    dev::eth::TransactionReceipts legacyPartialTransactionReceipts;
+    if ( m_db_ptr ) {
+        auto rawTransactionReceipts = m_db_ptr->getLegacyPartialTransactionReceipts();
+        if ( !rawTransactionReceipts.empty() ) {
+            dev::RLP rlp( rawTransactionReceipts );
+            dev::eth::BlockReceipts blockReceipts( rlp );
+            legacyPartialTransactionReceipts.insert( legacyPartialTransactionReceipts.end(),
```
olehnikolaiev: you can simply do `return blockReceipts.receipts;`

libethereum/Block.cpp
```diff
@@ -587,6 +591,16 @@ tuple< TransactionReceipts, unsigned > Block::syncEveryone( BlockChain const& _b
 
     LDB_CHECK( receipts.size() >= countBad );
 
+    if ( ClearPartialReceiptsPatch::isEnabledWhen( _timestamp ) ) {
+        // Making sure the old record was removed from the state db
+        m_state.safeRemoveLegacyPartialTransactionReceipts();
```
olehnikolaiev: can we optimize this process so that we won't have to try to remove non-existing key? for example, we can check the timestamp of the previous block and compare it with patchTimestamp

DmytroNazarenko (dismissed):

libethereum/Block.cpp
badrogger: addressed

libskale/State.cpp
badrogger: addressed

olehnikolaiev (approved):

DmytroNazarenko (approved):

PR #2175, opened 2025-05-12
Encrypt tx 'to' field

### LINKED ISSUE #2179: Encrypt `tx.to` field
### Description

- Encrypt `tx.to` field
- Original `tx.to` field should now be part of `data` field.
- Update `skaled_getTransactionData` call to return both decrypted data and `tx.to` fields.

PropzSaladaz:
## Description

- `to` field now goes inside data field, and also gets encrypted
- No longer use a prefix `BITE_MAGIC_NUMBER`
- We now use a `BITE_ADDRESS` of 20 bytes - size of an address, that replaces the original `to` field. Used to signal skaled that a transaction is encrypted using BITE protocol.
- Updated transaction receipts to include the decrypted `to` field.
- Updated `skaled_getTransactionData` call to return a JSON with 2 fields: 
   ```json
   {
     "data": ...
     "to": ...
   }
   ``` 
   
### Minor Updates

- Moved from using `ThresholdUtils` encryption / decryption functionality to use the more mature `ThresholdEncryption` functionality, which abstracts the `RANDOM_SECRET`, and also performs some additional tests on the ciphertext
- Transactions now get their BITE decrypted data as null when they are built. Then, consensus only passes the data of decrypted txs. Skaled can easily check if tx is BITE and data is not set -> then it was not successfully decrypted
- Added a script to ease local testing - that generates & imports ECDSA & BLS keys to sgxwallet & updated the config.json file with these new keys & runs skaled.

## Tests

- Modified all skaled tests to encrypt with new format, and check the decrypted `to` address + `data`
- Updated skaled tests to also check for the decrypted `to` field in transaction receipts
- Tested locally using sgx wallet, and `bite-ts` sending an encrypted transaction in the new format
- Tested on devnet:
   - Sending an encrypted transaction with `bite-ts`, and awaiting for the receipt with the `to` field decrypted
   - Sending `skale_getDecryptedTransactionData` and receiving both `data` and `to` fields decrypted

Part of skalenetwork/skale-consensus#877


libhistoric/AlethExecutive.cpp
```diff
@@ -130,16 +130,18 @@ bool AlethExecutive::execute() {
 
 #ifdef BITE
     bytes const& dataToPassToEvm = m_t.decryptedData();
+    Address const& receiverAddressToPassToEvm = m_t.decryptedTo();
```
olehnikolaiev: please use `Address` instead of `Address const&` as the temporary object it is referencing to might be deleted

libethcore/TransactionBase.h
```diff
@@ -149,22 +149,28 @@ class TransactionBase {
...
-    void setDecryptedData( const std::shared_ptr< bytes >& _decryptedData ) {
-        if ( _decryptedData )
+
+    void setDecryptedFields( const std::shared_ptr< bytes >& _decryptedData,
+        const std::shared_ptr< Address >& _decryptedTo ) {
+        if ( _decryptedData && _decryptedTo ) {
             m_decryptedData = _decryptedData;
-        m_isBITETxn = true;
+            m_decryptedTo = _decryptedTo;
+        }
```
olehnikolaiev, line 158: please set `m_isBITETxn` to `true` here. otherwise `isInvalidBiteTransaction()` always returns `false` for transactions that come from consensus

libethereum/Executive.cpp
```diff
@@ -293,15 +293,17 @@ bool Executive::execute() {
     // for BITE transactions returns decrypted data
     // for regular transactions returns regular data
     bytes const& dataToPassToEvm = m_t.decryptedData();
+    Address const& receiverAddressToPassToEvm = m_t.decryptedTo();
```
olehnikolaiev: please change `Address const&` to `Address`

libethereum/SkaleHost.cpp
```diff
@@ -592,12 +588,14 @@ void SkaleHost::createBlock( const ConsensusExtFace::transactions_vector& _appro
...
 #ifdef BITE
-            auto it = _decryptedTransactionDataFields->find( i );
-            if ( it != _decryptedTransactionDataFields->end() ) {
-                t.setDecryptedData( it->second );
-                if ( it->second )
-                    validDecryptedTransactionDataFields->insert( *it );
+            auto it = _decryptedTransactionFields->find( i );
+            if ( it != _decryptedTransactionFields->end() ) {
+                DecryptedTransactionFields& txFields = it->second;
+
```
olehnikolaiev, line 594: we should allow non-decryptable transactions to pass consensus stage and arrive to skaled. skaled then charges users who submitted those transactions for submitting invalid transactions on purpose

libethcore/TransactionBase.h
PropzSaladaz: That field is set upon transaction construction, so at this stage it should have been set already

kladkogex (dismissed):

DmytroNazarenko (approved):

PR #2189, opened 2025-05-29
Disable PoW for mirage

badrogger:
## Description
- Disable PoW functionality for MIRAGE

## Tests
- Fixed unit tests

## Performance Impact
- No performance related changes were made


olehnikolaiev (commented):
should we remove `externalGasDifficulty` from config? it is not used anywhere but in `checkOutExternalGas`

kladkogex (dismissed):

libethereum/ClientBase.cpp
```diff
@@ -120,13 +122,18 @@ std::pair< u256, ExecutionResult > ClientBase::estimateGas( Address const& _from
         if ( upperBound == Invalid256 || upperBound > c_maxGasEstimate )
             upperBound = c_maxGasEstimate;
         int64_t lowerBound;
+#ifdef MIRAGE
```
DmytroNazarenko: Why do we need this ifdef for MIRAGE?

DmytroNazarenko (changes requested):
Let's remove externalGasDifficulty from mirage build

DmytroNazarenko (approved):

olehnikolaiev (approved):

PR #2191, opened 2025-06-03
Disable IMA for MIRAGE

badrogger:
## Description

- Added disabling IMA functionality for MIRAGE
- Replaced `skale_imaInfo` API with `skale_getBLSPublicKey` for FAIR build:

`skale_getBLSPublicKey` is a JSON-RPC API that returns BLS Public Key of the node. The response format is as follows:

```json
{
  "BLSPublicKey0": "...",
  "BLSPublicKey1": "...",
  "BLSPublicKey2": "...",
  "BLSPublicKey3": "..."
}
```

## Tests

- Fixed unit tests
- Tested locally 
- Added unit test
- Tested on devnet

## Performance Impact

- No performance related changes were made


libethereum/ValidationSchemes.cpp
```diff
@@ -211,13 +215,17 @@ void validateConfigJson( js::mObject const& _obj ) {
             { "log-value-size-limit", { { js::int_type }, JsonFieldPresence::Optional } },
             { "log-json-string-limit", { { js::int_type }, JsonFieldPresence::Optional } },
             { "log-tx-params-limit", { { js::int_type }, JsonFieldPresence::Optional } },
+#ifndef MIRAGE
             { "no-ima-signing", { { js::bool_type }, JsonFieldPresence::Optional } },
+#endif
             { "skale-manager", { { js::obj_type }, JsonFieldPresence::Optional } },
```
olehnikolaiev: what about skale-manager` variable?

libethereum/Precompiled.cpp
```diff
@@ -699,7 +699,10 @@ ETH_REGISTER_PRECOMPILED( logTextMessage )( bytesConstRef _in ) {
 }
 
 static const std::list< std::string > g_listReadableConfigParts{ "skaleConfig.sChain.nodes.",
```
olehnikolaiev, line 705: I think this variable was required to use in precompiled contracts that we have disabled already (like `getConfigVariableU256` and etc). so we don't need it anymore

libskale/SnapshotHashAgent.cpp
```diff
@@ -256,8 +256,14 @@ std::tuple< dev::h256, libff::alt_bn128_G1, libff::alt_bn128_G2 > SnapshotHashAg
 
         libff::alt_bn128_G2 publicKey;
         if ( urlToDownloadSnapshotFrom_.empty() ) {
+#ifdef MIRAGE
+            auto publicKeyFromParams = chainParams_.sChain.currentGroups.back().BLSPublicKeys;
```
olehnikolaiev: here you should you individual bls public key instead of common bls public key. so we need to either ask a node via json rpc for its key or guess it from the config file that we have. it is also weird that this patch has passed the tests, lets take a look at them as well.

kladkogex (dismissed):

DmytroNazarenko (dismissed):

kladkogex (dismissed):

olehnikolaiev (approved):

PR #2210, opened 2025-06-17
Disable pre EIP-155 transactions

### LINKED ISSUE #2190: Disable Pre-EIP-155 transactions in Mirage
pre-EIP-155 transactions should lead to an error in transaction acceptance eth_sendRawTransaction and transaction broadcast

PropzSaladaz:
## Description

- Added new `bool` config variable `allowPreEIP155Txns`
    - If omitted, skaled allows pre-EIP155 txns by default (leave old tests' transaction validation logic unchanged)
    - If `true`, then allow pre-EIP155. Else, throws exception when validating received transaction via `eth_sendRawTransaction` (we should set it to `false` for FAIR builds)
    - The reason for it is to still be compatible with all older tests (avoid re-generating all the tests, with new hashes, signatures, and so on. Also avoids having dedicated MIRAGE test files - would result in double the files).
    - if  `allowPreEIP155Txns` is set to `true`, pre-EIP155 txs are rejected even if `disableCheckChainId` variable in config is set.
  
## Tests

- Added 3 tests to test for tx accceptance vs. rejection depending on the value of `allowPreEIP155Txns` from the config file

---

### Note
- This PR started by rejecting all pre-EIP155 txs by default - Transaction objects always needed chainId field, signature validation also rejected all pre-EIP155 signatures, etc. 
- Mid-implementation, it was decided that it would be much faster to add config variable, and not touch all older tests.
- Some changes of the initial implementation were left:
    - Added transaction constructors that receive `chainId` field. Currently not used, but may be useful in the future, if we decide reject pre-EIP155 at the core of the structs (Transaction, signature validation, etc).

Fixes #2190


libethcore/TransactionBase.cpp
```diff
@@ -534,14 +543,13 @@ void TransactionBase::checkLowS() const {
         BOOST_THROW_EXCEPTION( InvalidSignature() );
 }
 
-void TransactionBase::checkChainId( uint64_t chainId, bool disableChainIdCheck ) const {
-    if ( !disableChainIdCheck ) {
-        if ( !m_chainId.has_value() ) {
-            BOOST_THROW_EXCEPTION( InvalidTransactionFormat() );
-        }
+void TransactionBase::checkChainId( uint64_t chainId ) const {
```
kladkogex, line 546: It should be _chainId

badrogger (dismissed):

kladkogex (approved):

olehnikolaiev (approved):

PR #2245, opened 2025-07-30
2237 Download snapshot only if datadir is empty

### LINKED ISSUE #2237: ignore `--download-snapshot` option if data dir is not empty
if `--download-snapshot` is passed to skaled as a command-line it should be ignored if skaled has at least one block saved in the database. 

if snapshot downloading procedure is triggered by `skale-admin` it erases data dir by itself
if snapshot downloading procedure is triggered by `skaled`, `skale-admin` erases data dir as well

thus if skaled has non-empty data dir, snapshot downloading procedure should not be started

badrogger:
## Description

- To make sure skaled works properly after restart condition was added to download snapshot only if datadir is empty regardless of `--download-snapshot` flag

## Tests

- Tested on local network 
- Added unit test for isDataDirEmpty

## Performance Impact

- No performance related changes were introduced


libdevcore/FileSystem.cpp
```diff
@@ -71,6 +71,26 @@ fs::path dev::getDataDir( string _prefix ) {
...
+    }
+
+    if ( !fs::is_directory( dataDir ) ) {
+        return false;
+    }
+
+    try {
+        fs::directory_iterator begin( dataDir );
+        fs::directory_iterator end;
+        return begin == end;
```
DmytroNazarenko: Consider using fs::is_empty

skaled/main.cpp
```diff
@@ -619,6 +672,66 @@ void downloadAndProccessSnapshot( std::shared_ptr< SnapshotManager >& snapshotMa
...
+    std::shared_ptr< SnapshotManager >& snapshotManager,
+    std::shared_ptr< SharedSpace >& sharedSpace, bool zeroSnapshotOnly = false ) {
+    static Logger loggerInfo{ createLogger( VerbosityInfo, "doSnapshotDownload" ) };
+    static Logger loggerWarning{ createLogger( VerbosityWarning, "doSnapshotDownload" ) };
+#ifdef MIRAGE
+    // To process correct signatures we fetch current block timestamp
+    // from one of the nodes and temporarily changing current group
+
+    CurrentGroup latestGroup = chainParams->getNewestGroup();
+    uint64_t fetchedCurrentBlockTimetamp = fetchLatestBlockTimestampFromNodes( latestGroup.nodes );
```
DmytroNazarenko: Please, fix the typo

libdevcore/FileSystem.cpp
```diff
@@ -71,6 +71,26 @@ fs::path dev::getDataDir( string _prefix ) {
     return getDefaultDataDir( _prefix );
 }
 
+bool dev::isDataDirEmpty( string _prefix ) {
```
DmytroNazarenko, line 74: Let's pass `const std::string& _prefix` instead

DmytroNazarenko (changes requested):

libdevcore/FileSystem.cpp
badrogger: addressed

skaled/main.cpp
badrogger: addressed

DmytroNazarenko (dismissed):

DmytroNazarenko (approved):

PR #2266, opened 2025-09-02
build historic container optionally

### LINKED ISSUE #2267: dont build historic container for custom build by default
building historic container by default for custom build workflow is overhead because most of the time it is not needed and only increases the waiting time for the job to complete

olehnikolaiev:
running both build in parallel should be optional

### DIFF
--- a/.github/workflows/custom_build.yml
+++ b/.github/workflows/custom_build.yml
@@ -26,10 +26,7 @@ on:
         default: RelWithDebInfo
       run_historic_build:
         description: 'Run historic build'
-        type: choice
-        options:
-          - true
-          - false
+        type: boolean
         required: false
         default: false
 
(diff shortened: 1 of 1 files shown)


.github/workflows/custom_build.yml
```diff
@@ -24,6 +24,11 @@ on:
         - RelWithDebInfo
         - Release
         default: RelWithDebInfo
+      run_historic_build:
+        description: 'Run historic build'
+        type: boolean
+        required: false
+        default: false
```
DmytroNazarenko: Suggest adding an option with three possible values instead: core, historic, or both (default: both).

DmytroNazarenko (approved):

kladkogex (approved):

PR #2282, opened 2025-09-23
Bug/2280 wrong group index on start

### LINKED ISSUE #2280: skaled starts with wrong group index if exits right after rotation
if skaled exits right after rotation (committee or node) is completed, then on start it initializes with wrong group index. this leads to incorrect behavior and blockchain divergence 

this bug is the reason why https://github.com/skalenetwork/skale-consensus/issues/935 occured

olehnikolaiev:
fixes https://github.com/skalenetwork/skaled/issues/2280

if skaled exited right after the rotation was completed, then on startup it initialised with wrong group index. 

for L2 the bug was silent because the scenario to reproduce it is hard to make - the fix requires patch timestamp to make sure the existing chains won't be affected
for FAIR patch timestamp will be enabled by default

tested on local machine - bug was present for both L2 and FAIR builds

### DIFF
--- a/libethereum/SchainPatch.h
+++ b/libethereum/SchainPatch.h
@@ -181,6 +181,7 @@ DEFINE_SIMPLE_PATCH( CurrentBlockRandomPatch );
  * Context: fix group index initialization on startup
  * if skaled exits after the first block after rotation timestamp
  * then it starts again and initializes with wrong group index
+ * Version introduced: 4.1.0
  */
 DEFINE_AMNESIC_PATCH( GroupIndexInitPatch );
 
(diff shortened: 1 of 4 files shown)


kladkogex (dismissed):

libethereum/SchainPatch.h
```diff
@@ -177,4 +177,11 @@ DEFINE_SIMPLE_PATCH( InvalidTransactionFormatPatch );
  */
 DEFINE_SIMPLE_PATCH( CurrentBlockRandomPatch );
 
+/*
+ * Context: fix group index initialization on startup
+ * if skaled exits after the first block after rotation timestamp
+ * then it starts again and initializes with wrong group index
```
DmytroNazarenko, line 183: Add version please

badrogger (dismissed):

DmytroNazarenko (changes requested):

DmytroNazarenko (approved):

PR #2314, opened 2025-10-31
#2311 set min gas price via config

### LINKED ISSUE #2311: Add ability to set gasPrice from the config
Currently default minimum gasPrice is set as 100000 wei in consensus:
https://github.com/skalenetwork/skale-consensus/blob/03aed6804736fbbd5bcd26eeea818aa306b54c72/pricing/PricingAgent.cpp#L55

We need flexibility to configure the minimum gas price rather than relying on a static default. We need to allow passing a DYNAMIC_PRICING_MIN_PRICE parameter to consensus to set a custom minimum gasPrice.

olehnikolaiev:
fixes https://github.com/skalenetwork/skaled/issues/2311

add an option to set minimum gas price from config

verified on local machine

use the following to set minGasPrice to 123456 Wei
```
"skaleConfig": {
        "nodeInfo": {
            "dynamicPricingMinPrice": 123456,
            "dynamicPricingStartPrice": 123456,
```

### DIFF
--- a/libethereum/ValidationSchemes.cpp
+++ b/libethereum/ValidationSchemes.cpp
@@ -231,7 +231,8 @@ void validateConfigJson( js::mObject const& _obj ) {
             { "syncNodeCatchupTimeoutSec", { { js::int_type }, JsonFieldPresence::Optional } },
             { "readJsonHeaderTimeoutSec", { { js::int_type }, JsonFieldPresence::Optional } },
             { "syncNodeReadJsonHeaderTimeoutSec",
-                { { js::int_type }, JsonFieldPresence::Optional } } } );
+                { { js::int_type }, JsonFieldPresence::Optional } },
+            { "DYNAMIC_PRICING_MIN_PRICE", { { js::int_type }, JsonFieldPresence::Optional } } } );
 
 #ifndef FAIR
     std::string keyShareName = "";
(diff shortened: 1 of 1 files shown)


libethereum/ValidationSchemes.cpp
```diff
@@ -231,7 +231,8 @@ void validateConfigJson( js::mObject const& _obj ) {
             { "syncNodeCatchupTimeoutSec", { { js::int_type }, JsonFieldPresence::Optional } },
             { "readJsonHeaderTimeoutSec", { { js::int_type }, JsonFieldPresence::Optional } },
             { "syncNodeReadJsonHeaderTimeoutSec",
-                { { js::int_type }, JsonFieldPresence::Optional } } } );
+                { { js::int_type }, JsonFieldPresence::Optional } },
+            { "DYNAMIC_PRICING_MIN_PRICE", { { js::int_type }, JsonFieldPresence::Optional } } } );
```
DmytroNazarenko: We should have the same standard for config variables, please use camel case instead

DmytroNazarenko (changes requested):
Please, add an example of the config section with the new parameter in the PR description

kladkogex (dismissed):

DmytroNazarenko (approved):

PR #2320, opened 2025-11-12
IS-1341 fix block random

olehnikolaiev:
fixes https://github.com/skalenetwork/internal-support/issues/1341

pass `PrecompiledCallContext` to precompiled contracts to differentiate external call against transaction execution
`PrecompiledCallContext` can also be used in the future to pass more info to precompiled contracts

verified on a devnet

### DIFF
--- a/libethereum/Precompiled.h
+++ b/libethereum/Precompiled.h
@@ -51,7 +51,8 @@ extern std::shared_ptr< SkaleHost > g_skaleHost;
 
 struct PrecompiledCallContext {
     PrecompiledCallContext() : blockNumber( 0 ), isReadOnly( true ) {}
-    PrecompiledCallContext( dev::u256 bn, bool readOnly ) : blockNumber( bn ), isReadOnly( readOnly ) {}
+    PrecompiledCallContext( dev::u256 bn, bool readOnly )
+        : blockNumber( bn ), isReadOnly( readOnly ) {}
     dev::u256 blockNumber;
     bool isReadOnly;
 };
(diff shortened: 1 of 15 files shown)


badrogger (dismissed):

PropzSaladaz (dismissed):

libethereum/Precompiled.h
DmytroNazarenko: We can remove this

libethereum/SkaleHost.cpp
```diff
@@ -984,10 +984,12 @@ u256 SkaleHost::getGasPrice( unsigned _blockNumber ) const {
...
-        if ( _blockNumber > m_client.number() )
-            // cant get info about future blocks
-            // only possible if responding eth_call or eth_estimateGas request
-            _blockNumber = m_client.number();
+        if ( dev::eth::g_currentTransactionIndex == -1 ) {
+            // means a call outside of block is being executed
+            // if _blockNumber > currentBlockNumber, need to decrease it by 1
+            if ( _blockNumber > m_client.number() )
+                --_blockNumber;
+        }
```
DmytroNazarenko: Will it work correctly for historic node?

DmytroNazarenko (changes requested):

libethereum/SkaleHost.cpp
olehnikolaiev: if its a call to latest block - it works the same for core and historic node
if its a call to historic block - this condition is not hold and `m_consensus->getRandomForBlockId( _blockNumber )` is called

DmytroNazarenko: What if the call will be to block 0?

libskale/State.cpp
```diff
@@ -1114,6 +1114,7 @@ std::pair< ExecutionResult, TransactionReceipt > State::execute( EnvInfo const&
     // transaction is bad in any way.
```
kladkogex, line 1114: I think we need to remove these weird comments :)))

kladkogex (dismissed):

PropzSaladaz (approved):

kladkogex (approved):

badrogger (approved):

DmytroNazarenko (approved):

PR #2323, opened 2025-11-19
IS-1341 fix block random

olehnikolaiev:
fixes https://github.com/skalenetwork/internal-support/issues/1341

fix `getBlockRandom` precompiled contract execution for historic nodes
add tests

### DIFF
--- a/libethereum/SkaleHost.cpp
+++ b/libethereum/SkaleHost.cpp
@@ -982,11 +982,25 @@ u256 SkaleHost::getGasPrice( unsigned _blockNumber ) const {
     return m_consensus->getPriceForBlockId( _blockNumber );
 }
 
-u256 SkaleHost::getBlockRandom( unsigned _blockNumber ) const {
-    if ( CurrentBlockRandomPatch::isEnabledInWorkingBlock() ) {
+u256 SkaleHost::getBlockRandom( unsigned _blockNumber, bool _isCalledFromTxn ) const {
+    // for FAIR patch is always enabled
+    // check that patch enabled after block _blockNumber - 1
+    // if so - return correct value
+    // if not - return value for previous block
+    // works for historic calls, eth_call, eth_estimateGas
+    // and regular transactions
+    if ( CurrentBlockRandomPatch::isEnabledWhen(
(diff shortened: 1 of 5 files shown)


libethereum/SkaleHost.cpp
```diff
@@ -982,11 +982,25 @@ u256 SkaleHost::getGasPrice( unsigned _blockNumber ) const {
     return m_consensus->getPriceForBlockId( _blockNumber );
 }
 
-u256 SkaleHost::getBlockRandom( unsigned _blockNumber ) const {
-    if ( CurrentBlockRandomPatch::isEnabledInWorkingBlock() ) {
+u256 SkaleHost::getBlockRandom( unsigned _blockNumber, bool _isCalledFromTxn ) const {
```
DmytroNazarenko, line 985: What is the expected behaviour for _blockNumber = 0 ?

libethereum/SkaleHost.cpp
```diff
@@ -982,11 +982,25 @@ u256 SkaleHost::getGasPrice( unsigned _blockNumber ) const {
...
+    // for FAIR patch is always enabled
+    // check that patch enabled after block _blockNumber - 1
+    // if so - return correct value
+    // if not - return value for previous block
+    // works for historic calls, eth_call, eth_estimateGas
+    // and regular transactions
+    if ( CurrentBlockRandomPatch::isEnabledWhen(
+             m_client.blockInfo( m_client.hashFromNumber( _blockNumber - 1 ) ).timestamp() ) ) {
```
DmytroNazarenko: Suggest to retrieve this into a separate variable to increase readability

libethereum/SkaleHost.cpp
olehnikolaiev: it's a corner case, that should be handled separately. in general, no transactions are added to 0 block, so no consistency issues

libethereum/SkaleHost.cpp
olehnikolaiev: fixed

DmytroNazarenko (approved):

kladkogex (approved):

PR #2325, opened 2025-11-26
Feature/2297 bite2 create ctx

### LINKED ISSUE #2297: create BITE Phase 2 precompiled contract
create a precompiled contract `decryptAndExecute` for BITE Phase 2. this contract should:
- take 2 parameters as inputs - `bytes[] encryptedArguments`, `bytes[] plaintextArguments`
- creates a transaction with `m_data = functionSelector + RLP(encryptedArguments,plaintextArguments)`, `from=originalSenderAddress`, `to=contractCallerAddress` and inserts it into txn queue using `insertCAT(const Transaction& _cat)` method
- price for executing this precompiled contract should be a linear function of the size of `encryptedArguments` (aka `P * N + S`, where `N` - number of encrypted arguments, `P` - price for decrypting 1 argument, `S` - price for inserting a txn into txn queue)

olehnikolaiev:
fixes #2297 

add `submitCTX` precompiled smart contract to be called from external smart contract. `submitCTX` receives 4 arguments encoded using `abi.encode` - `bytes walletAndSignature, address destination, uint256 gasLimit, bytes data`. where `walletAndSignature` is the result of executing `getRandomWalletAndSignatureForCTX` precompiled and `data` is encoded using `abi.encode(bytes[] encryptedArgs, bytes[] plaintextArgs)`

modify`getRandomWalletAndSignatureForCTX` precompiled to return signature for BITE2 txn alongside generated wallet's address to not run complex math calculations twice with the same data

add `bite2TxnQueue` to push BITE2 txns there. BITE2 txns can only be created during transaction execution. `bite2TxnQueue` is called for pending txns after block processing is finished, so there is no need in synchronization. `bite2TxnQueue` is never flushed for now - this will be added in the upcoming prs :)    

add more tests


libethcore/TransactionBase.h
DmytroNazarenko: This include is used only for BITE2

libethereum/Precompiled.cpp
DmytroNazarenko, line 46: This include should be used only for BITE2

libethereum/Precompiled.cpp
```diff
@@ -29,6 +29,7 @@
 #include <libdevcore/CommonJS.h>
 #include <libdevcore/FileSystem.h>
 #include <libdevcore/Log.h>
+#include <libdevcore/RLP.h>
```
DmytroNazarenko, line 37: This include should be used only for BITE2

libethereum/Precompiled.cpp
DmytroNazarenko, line 1: `#include <libdevcore/Address.h>` in Precompiled.h should be used only for BITE2

DmytroNazarenko (changes requested):

libethereum/Precompiled.cpp
```diff
@@ -1092,7 +1093,228 @@ ETH_REGISTER_PRECOMPILED( getBlockRandom )( bytesConstRef, const PrecompiledCall
 
 #ifdef BITE2
 
-ETH_REGISTER_PRECOMPILED( getRandomWalletForCTX )
+dev::bytes abiEncodedArraysToRlp( const dev::bytes& _abiEncodedArrays ) {
```
DmytroNazarenko: Suggest to move it out of Precompiled.cpp

libethereum/Precompiled.cpp
DmytroNazarenko, line 1: Suggest to move error codes for BITE2 Precompiled contracts to a separate struct

libethereum/Precompiled.cpp
```diff
@@ -1092,7 +1093,228 @@ ETH_REGISTER_PRECOMPILED( getBlockRandom )( bytesConstRef, const PrecompiledCall
...
+        dev::bytes walletAndSigBytes =
+            _in.cropped( walletAndSigOffset.convert_to< size_t >() + 32, 116 ).toBytes();
+
+        // Extract wallet address (first 20 bytes)
+        dev::Address walletAddress(
+            dev::bytes( walletAndSigBytes.begin(), walletAndSigBytes.begin() + 20 ) );
+        if ( walletAddress == dev::ZeroAddress )
+            return { false, toBigEndian( dev::u256( 7 ) ) };
+        // verify account is not active
+        if ( g_skaleHost->client().countAt( walletAddress ) > 0 )
```
DmytroNazarenko, line 999: How will it work with historic calls?

libethereum/Precompiled.cpp
```diff
@@ -1092,7 +1093,228 @@ ETH_REGISTER_PRECOMPILED( getBlockRandom )( bytesConstRef, const PrecompiledCall
...
+                << "\n";
+            return { false, toBigEndian( dev::u256( 13 ) ) };
+        } catch ( ... ) {
+            BOOST_LOG( getLogger( VerbosityError ) )
+                << "Unknown exception in precompiled/submitCTX/abiEncodedArraysToRlp()\n";
+            return { false, toBigEndian( dev::u256( 14 ) ) };
+        }
+
+        // Get gas price
+        dev::u256 gasPrice = g_skaleHost->getGasPrice();
```
DmytroNazarenko: How will it work with historic calls? Suggest to pass the block from PrecompiledCallContext

libethereum/Precompiled.cpp
```diff
@@ -1105,40 +1327,94 @@ ETH_REGISTER_PRECOMPILED( getRandomWalletForCTX )
...
         auto evmSchedule = g_skaleHost->client().evmSchedule();
-        if ( TransactionBase::baseGasRequired(
-                 false, dev::bytesConstRef( data.data(), data.size() ), evmSchedule, true ) > gas )
-            return { false, toBigEndian( dev::u256( 4 ) ) };
+        if ( TransactionBase::baseGasRequired( false,
+                 dev::bytesConstRef( rlpEncodedData.data(), rlpEncodedData.size() ), evmSchedule,
+                 true ) > gas )
+            return { false, toBigEndian( dev::u256( 7 ) ) };
 
         dev::u256 gasPrice = g_skaleHost->getGasPrice();
```
DmytroNazarenko: How will it work with historic calls? Suggest to pass the block from PrecompiledCallContext

libethcore/TransactionBase.h
olehnikolaiev: it is also used by regular BITE

libethereum/Precompiled.cpp
olehnikolaiev: not used at all, will delete

libethereum/Precompiled.cpp
olehnikolaiev: this precompiled (submitCTX) is not executed for historic calls / eth_call / eth_estimateGas

libethereum/Precompiled.cpp
olehnikolaiev: this precompiled (submitCTX) is not executed for historic calls / eth_call / eth_estimateGas

libethereum/Precompiled.cpp
olehnikolaiev: suggest to move all helpers used in this file to a separate file then. cleaning up the entire file is probably more effective

libethcore/TransactionBase.h
```diff
@@ -488,7 +488,7 @@ class TransactionBase {
 #ifdef BITE2
     // Solidity adds 12 left-padded zero bytes when encoding an address parameter in the ABI format.
     static constexpr uint64_t BITE2_INPUT_DATA_MIN_LEN =
-        12 + dev::Address::size + dev::h256::size + BITE_CIPHERTEXT_MIN_LEN;
+        12 + dev::Address::size + dev::h256::size * 3 + BITE_CIPHERTEXT_MIN_LEN;
```
PropzSaladaz: I think it would be good to also add a comment on why the 3 * dev::h256::size / the macro structure

libethereum/Precompiled.cpp
```diff
@@ -1105,40 +1327,94 @@ ETH_REGISTER_PRECOMPILED( getRandomWalletForCTX )
...
+        dev::bytes rlpEncodedData;
+        try {
+            rlpEncodedData = abiEncodedArraysToRlp( data );
+        } catch ( std::exception& ex ) {
+            std::string strError = ex.what();
+            if ( strError.empty() )
+                strError = "exception without description";
+            BOOST_LOG( getLogger( VerbosityError ) )
+                << "Exception in precompiled/getRandomWalletForCTX/abiEncodedArraysToRlp(): "
+                << strError << "\n";
```
PropzSaladaz, line 1153: The catch below has different behavior than this, as the one below returns with error. Shouldn't this also return with error?

PropzSaladaz (changes requested):

libethereum/Precompiled.cpp
olehnikolaiev: it's a typo, there should be return for both catches. will fix

libethereum/Precompiled.cpp
```diff
@@ -22,13 +22,18 @@
  */
 
 #include "Precompiled.h"
+#include "BITEConstants.h"
```
DmytroNazarenko, line 26: This include should be wrapped in `ifdef BITE2`

libethereum/BITEConstants.h
```diff
@@ -0,0 +1,61 @@
...
+// walletAndSignature = wallet_address(20) + r(32) + s(32) + v(32)
+static constexpr uint64_t BITE2_TRANSACTION_SUBMITION_INPUT_DATA_MIN_LEN =
+    BITE2_WALLET_GENERATION_INPUT_DATA_MIN_LEN + WALLET_AND_SIGNATURE_LENGTH + dev::h256::size;
+
+// keccak(onDecrypt(bytes[],bytes[])) - first 4 bytes: 0x57983ac8
+static inline const dev::bytes ON_DECRYPT_FUNCTION_SELECTOR = { 0x57, 0x98, 0x3a, 0xc8 };
+
+// Error codes for submitCTX precompiled contract
+namespace SubmitCTXErrors {
+constexpr uint64_t SUCCESS = 1;
```
DmytroNazarenko, line 30: Let's retrieve SUCCESS code or rename the namespace to avoid confusion

libethereum/TransactionQueue.cpp
```diff
@@ -565,3 +565,14 @@ Transactions TransactionQueue::debugGetFutureTransactions() const {
...
     return res;
 }
+
+#ifdef BITE2
+void TransactionQueue::importBITE2Transaction( Transaction&& _t ) {
+    BOOST_LOG( m_loggerTrace ) << "BITE2 txn arrived";
+    m_bite2Current.push_back( std::move( _t ) );
+}
+
+std::vector< Transaction > TransactionQueue::pendingBITE2Transactions() const {
```
DmytroNazarenko, line 575: Suggestion: create debug API to get pending BITE2 transactions

libethereum/PrecompiledHelpers.cpp
DmytroNazarenko, line 1: Propose to move it to `libdevcore`

libethereum/BITEConstants.h
DmytroNazarenko, line 1: Propose to move it to `libdevcore`

DmytroNazarenko (changes requested):

libethereum/TransactionQueue.cpp
olehnikolaiev: it will be done in the next pr

libethereum/PrecompiledHelpers.cpp
olehnikolaiev: I think as far as Precompiled stays in libethereum, PrecompiledHelpers should probably stay there as well

libethereum/BITEConstants.h
olehnikolaiev: I tried it initially - moving it to libdevcore creates circular dependencies between lower-level (libdevcore) and higher-level(libconsensus) libraries, which I couldn't resolve without decreasing code quality. for that reason I decided to keep it in libethereum for now.

DmytroNazarenko (dismissed):

libethereum/PrecompiledHelpers.cpp
```diff
@@ -0,0 +1,304 @@
...
+    return ret;
+}
+
+Logger& getLogger( int a_severity ) {
+    static std::mutex g_mtx;
+    std::lock_guard< std::mutex > lock( g_mtx );
+    typedef std::map< int, Logger > map_loggers_t;
+    static map_loggers_t g_mapLoggers;
+    if ( g_mapLoggers.find( a_severity ) == g_mapLoggers.end() )
+        g_mapLoggers[a_severity] = Logger( boost::log::keywords::severity = a_severity,
```
kladkogex, line 78: Please use .at() instead of  []

libethcore/TransactionBase.cpp
```diff
@@ -567,6 +572,13 @@ int64_t TransactionBase::baseGasRequired(
         g += _es.BITETxnCost;
 #endif
 
+#ifdef BITE2
+    // BITE2 transaction - charge for every encrypted payload
+    if ( _bite2EncryptedArgsSize.has_value() ) {
+        g += _bite2EncryptedArgsSize.value() * _es.BITETxnCost;
+    }
+#endif
```
kladkogex, line 594: Security : Please add a check that there is no u256 integer overflow during multiplication or addition

libethereum/PrecompiledHelpers.cpp
```diff
@@ -0,0 +1,304 @@
...
+}
+
+boost::filesystem::path getFileStorageDir( const Address& _address ) {
+    return dev::getDataDir() / "filestorage" / _address.hex();
+}
+
+static const std::list< std::string > g_listReadableConfigParts{ "skaleConfig.sChain.nodes.",
+    "skaleConfig.nodeInfo.wallets.ima.n" };
+
+bool stat_is_accessible_json_path( const std::string& strPath ) {
```
kladkogex: Please use camelCase for function names

PropzSaladaz (dismissed):

kladkogex (dismissed):

libethereum/PrecompiledHelpers.cpp
olehnikolaiev: here we create an element if it is missing, so usage of [] is correct

kladkogex (approved):

PropzSaladaz (approved):

PR #2329, opened 2025-12-11
1247 leveldb options

badrogger:
## Description

This pull request updates the LevelDB configuration in State.cpp to reduce the number of compactions.

The goal is to minimize the likelihood of invalid state writes under high load. By increasing certain memory-related parameters, more data can be held in memory, resulting in less frequent flushes and compactions.

Key configuration changes:

`write_buffer_size` – Increases the size of the in-memory MemTable, allowing more state data to be buffered before flushing, thereby reducing flush frequency.

`max_file_size` – Raises the maximum size of SSTable files (starting from level 1), which helps reduce the number of compactions.

`max_open_files` – Allows LevelDB to keep more files open simultaneously (before it was 1,000), improving performance without significantly impacting the rest of the system.

`block_size` - Helps to reduce memory consumption by moving more data to cache.  

## Target issue 
Should be applied to fix https://github.com/skalenetwork/internal-support/issues/1247.

## Tests

- Tested on local network.
- Tested on mainnet indexer node (nebula).

## Performance 
- Memory consumption increased (~80Mb) , because more data holds in memory (`write_buffer_size` increased), but plot below shows similar behavior overall. 
- Catchup has similar performance. 
- Disk read performance behaves similarly.
- Disk write curve has higher peaks, but difference is around ~20Mb.

#### Blocks

Pasted image 20251211170544

Pasted image 20251211160214

#### Leveldb 
Pasted image 20251211160250

#### OS

Pasted image 20251211160516
Pasted image 20251211160531
Pasted image 20251211160601
Pasted image 20251211160545

#### Blocks 
block-per-second

blocks-per-second

#### LevelDB
comp

#### OS
ram
reads

writes

cpu

### Steps to reproduce
The best way to reproduce the bug is to restore nebula indexer from snapshot.

### DIFF
--- a/libskale/State.cpp
+++ b/libskale/State.cpp
@@ -269,6 +269,7 @@ db::LevelDB::LevelDBOptions State::getLevelDBConfig() const {
     levelDbOptions.dbOptions.write_buffer_size = 64 * 1024 * 1024;
     levelDbOptions.dbOptions.max_file_size = 16 * 1024 * 1024;
     levelDbOptions.dbOptions.max_open_files = 2000;
+    levelDbOptions.dbOptions.block_size = 32 * 1024;
     return levelDbOptions;
 }
 
--- a/libskale/State.cpp
+++ b/libskale/State.cpp
@@ -240,7 +240,13 @@ skale::OverlayDB State::openDB(
 
     fs::path state_path = path / fs::path( "state" );
     try {
-        m_orig_db.reset( new db::DBImpl( state_path ) );
+        db::LevelDB::LevelDBOptions levelDbOptions;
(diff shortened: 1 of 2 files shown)


kladkogex (dismissed):

badrogger:
No excessive performance downgrade on 4-node network with 5 chain, with 7-10 transactions per block for each chain were noticed 
image
image
image

olehnikolaiev (approved):

badrogger:
The best way to reproduce the bug is to restore nebula indexer from snapshot.

DmytroNazarenko (approved):
Please, create a ticket related to historic_state.db changes

PR #2336, opened 2025-12-17
Feature/2299 import and execute bite2 txns

### LINKED ISSUE #2298: create BITE Phase 2 txn queue private calls
BITE Phase 2 needs specific private calls for skaled to store txns generated by precompiled SC and for consensus to fetch those transactions

`insertCAT(const Transaction& _cat)` - allows precompiled SC to insert CAT into txn queue
`pendingCATs()` - allows consensus to fetch all CATs generated in previous block

### LINKED ISSUE #2299: import BITE Phase 2 txns from consensus
skaled - consensus interface should be updated to 
```
    virtual transactions_vector ConsensusExtFace::pendingTransactions(size_t _limit, u256 &_stateRoot
#ifdef BITE2
    , transactions_vector& _catTransactions
#endif
    );
```
and support `catTransactions` in a block

### LINKED ISSUE #2326: add debug api call to get pending bite2 transactions
add debug api call to get pending bite2 transactions

olehnikolaiev:
fixes #2299 #2326 #2298 

- push BITE2 txns to consensus
- process BITE2 txns arrived from consensus
- handle BITE2 txns during txn parsing
- add separate debug API call to get list of pending BITE2 txns


libdevcore/CommonData.cpp
```diff
@@ -116,3 +117,107 @@ std::string dev::toString( string32 const& _s ) {
         ret.push_back( _s[i] );
     return ret;
 }
+
+#ifdef BITE2
+dev::bytes dev::encodeArray( const std::vector< dev::bytes >& _elements ) {
```
PropzSaladaz: maybe we should change method name to `abiEncodeArray`, to make the type of encoding explicit

kladkogex (dismissed):

libethcore/TransactionBase.cpp
```diff
@@ -379,7 +380,15 @@ TransactionBase::TransactionBase( bytesConstRef _rlpData, CheckTransaction _chec
         // bad formatted txns cannot make it to the block
         // therefore no need to check it anywhere else
         checkIfBITETxnAndSet( m_receiveAddress );
-#endif
+
+#ifdef BITE2
+        // check if a txn is a CTX here
+        // bad formatted txns cannot make it to the block
+        // therefore no need to check it anywhere else
+        checkIfCTXAndSet( m_data );
```
PropzSaladaz, line 388: Could there be a case where a BITE1 tx has its encrypted data field starting with the same signature as BITE2 tx?

In such case, the `checkIfBITETxnAndSet` would set it as bite, and afterwards, `checkIfCTXAndSet` would also check it as CTX.
Although it should be extremely rare to happen, maybe we should either add an if else to only allow 1 of the branches, or just move the CTX above the BITE, since BITE1 is less strict

libethereum/BlockChain.cpp
```diff
@@ -772,9 +773,9 @@ void BlockChain::insertTransactionsDetailsToDb(
         RLP txns_rlp = blockRLP[1];
 
 #ifdef BITE
-        auto decryptedTransactionFieldsIt = _block.decryptedTransactionDataFields->begin();
+        //        CHECK_EXPRESSION( _block.decryptedTransactions.regularTxsMap );
```
PropzSaladaz: I think we should uncomment this line

libethereum/SkaleHost.cpp
```diff
@@ -991,6 +967,129 @@ void SkaleHost::broadcastFunc() {
...
+        Transaction t( data, CheckTransaction::Everything, true,
+            EIP1559TransactionsPatch::isEnabledInWorkingBlock(),
+            InvalidTransactionFormatPatch::isEnabledInWorkingBlock() );
+#ifdef BITE
+        if ( regularTxnsIterator != _decryptedTransactions.regularTxsMap->end() &&
+             regularTxnsIterator->first == i ) {
+            std::optional< DecryptedRegularTxFields > txFields = regularTxnsIterator->second;
+            if ( txFields.has_value() ) {
+                dev::Address to( txFields->to.data(), dev::Address::ConstructFromPointer );
+                ;
```
PropzSaladaz: we can remove the `;`

PropzSaladaz (changes requested):

libethcore/TransactionBase.cpp
olehnikolaiev: Each encrypted transaction's data for BITE1 is RLP-encoded as `RLP([EPOCH_ID, ENCRYPTED_BITE_DATA])`, and RLP-encoding for arrays always starts with specific prefix (`0xc0` - `0xff`) which doesn't overlap with BITE2_FUNCTION_SELECTOR - `0x57983AC8`

PropzSaladaz (approved):

kladkogex (approved):

PR #2396, opened 2026-02-10
Feature/2385 bind txn origin to ctx

### LINKED ISSUE #2385: save CTXs crafted by specific transaction
if a particular transaction `T` crafted CTXs, the list of CTXs's hashes should be saved in database and be associated with `T`. this will improve UX and allow smooth integration with BITE2 protocol for external tools

olehnikolaiev:
fixes #2385 

add new db entry for every block with list of lists of CTX hashes created in this block by every transaction
add new rpc methods `bite_getCtxOrigin` and `bite_getCrfatedCtxs`

### DIFF
--- a/docs/features/json-rpc-interface.md
+++ b/docs/features/json-rpc-interface.md
@@ -1164,9 +1164,7 @@ Transaction hash: "0x"-prefixed hex `String`, 32 bytes
 
 #### Example
 ```json
-[
   "0xb217cdf70dd4691e53040030b8931807b02633d92052d8c74f38b20944788922"
-]
 ```
 
 #### Exceptions
--- a/docs/features/json-rpc-interface.md
+++ b/docs/features/json-rpc-interface.md
@@ -1131,6 +1131,49 @@ Object with the following fields:
 - Throws error if the transaction has not been successfully decrypted
 
 
+### `bite_getCraftedCtxs`
+#### Description
+Retrieves the list of hashes of CTXs crafted by given transaction.
+
+#### Parameters
+1. Transaction hash: "0x"-prefixed hex `String`, 32 bytes
+
+#### Return format
+A JSON array of transaction hashes, each "0x"-prefixed hex `String`, 32 bytes.
+
+#### Example
+```json
+[
+  "0xb217cdf70dd4691e53040030b8931807b02633d92052d8c74f38b20944788922"
+]
+```
+
+#### Exceptions
+- Throws error if the transaction hash does not exist
+
+
+### `bite_getCtxOrigin`
+#### Description
+Retrieves the transaction that created particular CTX.
+
+#### Parameters
+1. Transaction hash: "0x"-prefixed hex `String`, 32 bytes
+
+#### Return format
+Transaction hash: "0x"-prefixed hex `String`, 32 bytes
+
+#### Example
+```json
+[
+  "0xb217cdf70dd4691e53040030b8931807b02633d92052d8c74f38b20944788922"
+]
+```
+
+#### Exceptions
+- Throws error if the transaction hash does not exist
+- Throws error if given transaction is not CTX
+
+
 ## Non-standard Methods
 ### `oracle_submitRequest`
 ### `oracle_checkResult`
--- a/libethereum/BITE2TransactionQueue.h
+++ b/libethereum/BITE2TransactionQueue.h
@@ -61,7 +61,7 @@ public:
 
 private:
     std::vector< Transaction > m_current;
-    std::atomic_size_t m_currentHeadIndex = -1;
+    std::atomic_size_t m_currentHeadIndex = 0;
     bool m_empty = true;
     mutable SharedMutex m_lock;
 
--- a/libethereum/BITE2TransactionQueue.h
+++ b/libethereum/BITE2TransactionQueue.h
@@ -39,8 +39,9 @@ public:
     /// No lock, returns reference to entire buffer. For internal logic.
     const std::vector< Transaction >& pendingBITE2Transactions() const;
 
-    // addTemp(), commitTemp(), clearTemp(), clear() and finalize() are always called from the same
-    // thread and pendingBITE2Transactions() always called AFTER methods above are executed
+    // addTemp(), getTempHashes(), commitTemp(), clearTemp(), clear() and finalize()
+    // are always called from the same thread
+    // and pendingBITE2Transactions() always called AFTER methods above are executed
     // therefore, there is no need in extra synchronization
     // only debug_pendingBITE2Transactions requires synchronization
     // because it is used by JSON RPC API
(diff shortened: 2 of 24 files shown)


badrogger (dismissed):

libethereum/ClientBase.cpp
```diff
@@ -504,7 +504,29 @@ TransactionHashes ClientBase::transactionHashes( h256 _blockHash ) const {
 DecryptedTransactionData ClientBase::decryptedTransactionData( h256 _transactionHash ) const {
     return bc().decryptedTransactionData( _transactionHash );
 }
-#endif
+
+#ifdef BITE2
+dev::h256 ClientBase::ctxOrigin( const dev::h256& _ctxHash ) const {
+    auto tx = transaction( _ctxHash );
+    if ( !tx.isCTX() )
+        std::logic_error( "Trying to get ctxOrigin for non-CTX" );
```
PropzSaladaz: Missing `throw` in line 512.

Then on line 520 I think we should throw an error as well - this feels more like a 'I tried but didn't find any CTX with that hash in block ctxHashLists', which is a case that shouldn't happen

libethereum/BITE2TransactionQueue.h
```diff
@@ -58,7 +61,8 @@ class BITE2TransactionQueue {
 
 private:
     std::vector< Transaction > m_current;
-    std::atomic_int m_currentHeadIndex = -1;
+    std::atomic_size_t m_currentHeadIndex = -1;
```
PropzSaladaz: Should we initialize it to 0 instead, since now the type is unsigned?

libethereum/ClientBase.cpp
```diff
@@ -504,7 +504,29 @@ TransactionHashes ClientBase::transactionHashes( h256 _blockHash ) const {
...
+    if ( originIndex != size_t( -1 ) ) {
+        return bc().transactionHashes( prevBlock )[originIndex];
+    }
+    return dev::h256();
+}
+
+std::vector< dev::h256 > ClientBase::craftedCTXs( const dev::h256& _transactionHash ) const {
+    auto tl = bc().transactionLocation( _transactionHash );
+    CtxOrigin ctxHashesLists = bc().ctxHashesForBlock( tl.first );
+    return ctxHashesLists[tl.second];
```
PropzSaladaz, line 531: I think we should add bound checks, or use `at()`

libethereum/BlockDetails.h
```diff
@@ -179,7 +179,45 @@ struct DecryptedTransactionData {
...
+    }
+    explicit operator bool() const { return !m_ctxHashesLists.empty(); }
+    std::vector< dev::h256 > operator[]( size_t i ) const { return m_ctxHashesLists[i]; }
+    size_t find( const dev::h256& _ctxHash ) const {
+        for ( size_t i = 0; i < m_ctxHashesLists.size(); ++i ) {
+            auto it = std::find( m_ctxHashesLists[i].begin(), m_ctxHashesLists[i].end(), _ctxHash );
+            if ( it != m_ctxHashesLists[i].end() )
+                return i;
+        }
+        return -1;
```
PropzSaladaz: Maybe we can change return type to `optional<>`, and return `nullopt` instead of -1.
-1 for unsiged type may lead to unexpected situations

docs/features/json-rpc-interface.md
```diff
@@ -1131,6 +1131,49 @@ Object with the following fields:
...
+
+#### Parameters
+1. Transaction hash: "0x"-prefixed hex `String`, 32 bytes
+
+#### Return format
+Transaction hash: "0x"-prefixed hex `String`, 32 bytes
+
+#### Example
+```json
+[
```
PropzSaladaz: we should leave the example as simple string instead of array. Else, both examples show same return type, but specification says otherwise

libethereum/BlockChain.cpp
```diff
@@ -1539,9 +1566,13 @@ void BlockChain::clearCachesDuringChainReversion( unsigned _firstInvalid ) {
     m_transactionAddresses.clear();  // TODO: could perhaps delete them individually?
 
 #ifdef BITE
-    DEV_WRITE_GUARDED( x_decryptedTransactionsData )
+    DEV_WRITE_GUARDED( x_decryptedTransactionsData );
```
PropzSaladaz: Should we not set the `;` at the end? Else, the macro (runs a single iteration loop) will not run any code within the loop, meaning the following code in line 1570 is not mutex guarded?
Same for line 1572

libweb3jsonrpc/SkaleFace.h
```diff
@@ -186,7 +198,18 @@ class SkaleFace : public ServerInterface< SkaleFace > {
...
             &dev::rpc::SkaleFace::bite_getDecryptedTransactionDataI );
-#endif
+#ifdef BITE2
+        this->bindAndAddMethod(
+            jsonrpc::Procedure( "bite_getCraftedCtxs", jsonrpc::PARAMS_BY_POSITION,
+                jsonrpc::JSON_ARRAY, "param1", jsonrpc::JSON_STRING, NULL ),
+            &dev::rpc::SkaleFace::bite_getCraftedCtxsI );
+        this->bindAndAddMethod(
+            jsonrpc::Procedure( "bite_getCtxOrigin", jsonrpc::PARAMS_BY_POSITION,
+                jsonrpc::JSON_OBJECT, "param1", jsonrpc::JSON_STRING, NULL ),
```
PropzSaladaz: Should we set JSON_STRING as return type?

libethereum/ClientBase.cpp
olehnikolaiev: line 512 fixed
as for the second part - usually when the requested item is not found in the db, this class returns default object instead of throwing an error.

libethereum/BITE2TransactionQueue.h
olehnikolaiev: fixed

docs/features/json-rpc-interface.md
olehnikolaiev: fixed

libweb3jsonrpc/SkaleFace.h
olehnikolaiev: fixed

libethereum/BlockChain.cpp
olehnikolaiev: fixed

libethereum/ClientBase.cpp
olehnikolaiev: fixed

libethereum/BlockDetails.h
olehnikolaiev: fixed

PropzSaladaz (approved):

badrogger (approved):

kladkogex (approved):

PR #2457, opened 2026-03-29
Bug/is 1410 ctx gas limit

olehnikolaiev:
fixes https://github.com/skalenetwork/internal-support/issues/1410

set max `gasLimit` for `CTX` created via `submitCTX` precompiled contract
disable `POW` calculation for `CTXs`

### DIFF
--- a/libethereum/Precompiled.cpp
+++ b/libethereum/Precompiled.cpp
@@ -957,8 +957,7 @@ ETH_REGISTER_PRECOMPILED( submitCTX )( bytesConstRef _in, const PrecompiledCallC
 
         // Extract gas limit from first 32 bytes
         bigint const gas( parseBigEndianRightPadded( _in, 0, dev::h256::size ) );
-        if ( gas <= 0 ||
-             gas > g_skaleHost->client().chainParams().getGasLimit() )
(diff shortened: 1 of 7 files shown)


libethereum/Precompiled.cpp
```diff
@@ -957,7 +957,7 @@ ETH_REGISTER_PRECOMPILED( submitCTX )( bytesConstRef _in, const PrecompiledCallC
 
         // Extract gas limit from first 32 bytes
         bigint const gas( parseBigEndianRightPadded( _in, 0, dev::h256::size ) );
-        if ( gas <= 0 )
+        if ( gas <= 0 || gas > g_skaleHost->client().chainParams().getGasLimit() )
```
DmytroNazarenko, line 960: How will this work for historical blocks? If skaled fetches block N where the blockGasLimit X exceeds the default blockGasLimit (and contains CTX with gasLimit X), this could cause the CTX creation to revert

DmytroNazarenko: The maximum gasLimit for any transaction type in skaled is capped by the default blockGasLimit (268M)

badrogger (dismissed):

kladkogex (approved):

DmytroNazarenko (approved):

PR #2481, opened 2026-06-01
Feature/2340 refund original ctx creator

### LINKED ISSUE #2340: refund the original CTX creator BITE2 temporary accounts after executing corresponding txns
currently we generate temporary accounts to send BITE2 txns. after corresponding BITE2 txn is executed the account may still have non-zero balance, but there is no entity that has access to it. suggest to burn the native tokens such accounts have left to avoid storage exhaustion and performance degradation

PropzSaladaz:


PropzSaladaz:
This PR will be closed. Solution for this issue was slightly changed and implemented in #2492

PR #2489, opened 2026-06-12
#2488 fix UB in tests

### LINKED ISSUE #2488: Non-deterministic failures in tests
## Description

2 non deterministic failures were identified that cause 2 tests to fail from time to time and slow down CI/CD:

### Issue 1

`SnapshotHashingTest` intermittently crashed due to a race condition between block import and snapshot creation:

* **The Problem:** The test triggered a snapshot as soon as `client->number()` changed. However, the block number updates *before* the full import flow completes. This allowed snapshots to run while live BTRFS/DB volumes were still being actively written to.

### Issue 2

In BITE builds, `submitCTX` tests intermittently crash during teardown due to an initialization/destruction ordering bug:

* **The Problem:** A local `Cleanup` struct nulled the global `g_skaleHost` *before* the `client` `unique_ptr` destroyed and joined its consensus worker thread (due to reverse-declaration order). This allowed a live worker thread to attempt block creation, unconditionally dereferencing `g_skaleHost` and causing a segmentation fault (`nullptr` offset read).

PropzSaladaz:
## Description

2 non deterministic failures were identified that cause 2 tests to fail from time to time and slow down CI/CD:

### Issue 1

`SnapshotHashingTest` intermittently crashed due to a race condition between block import and snapshot creation:

* **The Problem:** The test triggered a snapshot as soon as `client->number()` changed. However, the block number updates *before* the full import flow completes. This allowed snapshots to run while live BTRFS/DB volumes were still being actively written to.
* **The Fix:** Introduced `ClientTest::withBlockImportBarrier(...)`, which utilizes the existing `m_blockImportMutex`. Running snapshot creation under this barrier guarantees that no block import operations are modifying the underlying data at the same time.

### Issue 2

In BITE builds, `submitCTX` tests intermittently crash during teardown due to an initialization/destruction ordering bug:

* **The Problem:** A local `Cleanup` struct nulled the global `g_skaleHost` *before* the `client` `unique_ptr` destroyed and joined its consensus worker thread (due to reverse-declaration order). This allowed a live worker thread to attempt block creation, unconditionally dereferencing `g_skaleHost` and causing a segmentation fault (`nullptr` offset read).
* **The Fix:** Replaced `Cleanup` with a shared `BITEHostGuard` RAII struct. Its destructor explicitly stops/joins the worker thread via `stopWorking()` *before* clearing `g_skaleHost`, eliminating the race condition and deduplicating the teardown logic.

Fixes #2488


kladkogex (commented):
Approved

kladkogex (approved):

badrogger (approved):

